Issue #2, opened 2015-11-10
cf-release need to be checked out with `git submodule update --init --recursive` when using an absolute filepath


voelzmo:
A hint that your cf-release needs to have all recursive submodules would be appreciated. Maybe just referencing https://github.com/cloudfoundry/cf-release/blob/master/scripts/update as this is what your internal scripting calls when using `integration-latest` would be a good idea?

Amit-PivotalLabs:
Done: https://github.com/cloudfoundry/cf-deployment/commit/b26555201721d9a5291a466265af70ecec179fd7

PR #5, opened 2015-11-10
Make sure submodules are checked out recursively

voelzmo:
[Fixes https://github.com/cloudfoundry/cf-deployment/issues/2]


Amit-PivotalLabs:
The only issue with this is if you have work in flight, you don't want to run `./scripts/update` as it will blow away your changes.  How do we capture the idea that you essentially want to be in a `./scripts/update` state modulo any work-in-progress you may currently have in that directory or its submodules?

voelzmo:
Fair enough. Should we just mention then in the readme that you'll need all submodules checked out recursively? I don't know what the primary use-case for people here is. For you, it is probably making changes to cf-release and then deploying it. For others, it might to just deploy some version of cf.

Amit-PivotalLabs:
Done: https://github.com/cloudfoundry/cf-deployment/commit/b26555201721d9a5291a466265af70ecec179fd7

PR #16, opened 2016-10-29
Update service names for nats, mysql, cc_clock

mcwumbly:
Fresh deployments using `cf-deployment` were failing.

`metron_agent` logs on the consul instances pointed to issues reaching `etcd`.

Updating the service names below fixed the issue for us.

/cc @zankich @rosenhouse

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -89,7 +89,7 @@ instance_groups:
           domain: cf.internal
           servers: *3
           services:
-            etcd: {}
+            nats: {}
         encrypt_keys: "((consul_encrypt_keys))"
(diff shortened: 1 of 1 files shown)


dsabeti:
Thanks for the PR @mcwumbly. Thanks for bringing this to our attention. Plot twist: after looking into those instances, it turns out that none of those jobs should even be advertising themselves as services over consul. [We're just taking those out entirely](https://www.pivotaltracker.com/story/show/133457205). But thanks again for pointin this out, as it's prompting another survey of the whole manifest.

/cc @JesseTAlford

mcwumbly:
OK, great, we'll use `develop` until that's merged into `master`.

PR #18, opened 2016-11-02
Move static IPs to bosh vars

nelsam:
This moves the static IPs to variables, so that anyone with different subnets can alter them.  The change could potentially be a bit lighter, but this seems to work fine, and I didn't want to throw much more effort at it when it's just going to magically disappear with `bosh links`.

The old static IPs are now in default_networks.yml.  Merging this PR will likely need to be coordinated with any teams that already have a working CI with this manifest, as they'll need to update their `bosh deploy` step to include this file as a vars flag to the deploy command.

[#132676971]

Signed-off-by: Sam Nelson <snelson@pivotal.io>


rosenhouse:
I believe that this PR would make it possible to use `cf-deployment` with the `--ops-file`.  That would help our team move away from our current hack of maintaining [a patch against cf-deployment](https://github.com/cloudfoundry-incubator/container-networking-ci/blob/master/netman-cf-deployment.patch)

dsabeti:
Hey @nelsam and @rosenhouse. I'm thinking about merging this in, but I wanted to discuss other solutions first. What if instead, we provided a `static-ip-overrides.yml` ops file, that looks kind of like this:
```
# Replace the constants with the desired static IPs for each job
- type: replace
  path: '/instance_groups/name=nats/networks/name=private/static_ips'
  value: &nats_ips
  - NATS_STATIC_IP_1
  - NATS_STATIC_IP_2
- type: replace
  path: '/instance_groups/name=etcd/networks/name=private/static_ips'
  value: &etcd_static_ips
  - ETCD_STATIC_IP_1
  - ETCD_STATIC_IP_2
  - ETCD_STATIC_IP_3
- type: replace
  path: '/instance_groups/name=consul/networks/name=private/static_ips'
  value: &consul_static_ips
  - CONSUL_STATIC_IP_1
  - CONSUL_STATIC_IP_2
  - CONSUL_STATIC_IP_3
- type: replace
  path: '/instance_groups/name=mysql/networks/name=private/static_ips'
  value:
  - &mysql_static_ip MYSQL_STATIC_IP




### Do not edit below here
- type: replace
  path: '/instance_groups/name=nats/jobs/name=nats_stream_forwarder/properties/nats/machines'
  value: *nats_ips
- type: replace
  path: '/instance_groups/name=api/jobs/name=cloud_controller_ng/properties/nats/machines'
  value: *nats_ips
- type: replace
  path: '/instance_groups/name=api/jobs/name=cloud_controller_worker/properties/nats/machines'
  value: *nats_ips
- type: replace
  path: '/instance_groups/name=cc_clock/jobs/name=cloud_controller_clock/properties/nats/machines'
  value: *nats_ips
- type: replace
  path: '/instance_groups/name=router/jobs/name=gorouter/properties/nats/machines'
  value: *nats_ips
- type: replace
  path: '/instance_groups/name=route-emitter/jobs/name=route_emitter/properties/diego/route_emitter/nats/machines'
  value: *nats_ips
- type: replace
  path: '/instance_groups/name=api/jobs/name=route_registrar/properties/nats/machines'
  value: *nats_ips
- type: replace
  path: '/instance_groups/name=uaa/jobs/name=route_registrar/properties/nats/machines'
  value: *nats_ips
- type: replace
  path: '/instance_groups/name=blobstore/jobs/name=route_registrar/properties/nats/machines'
  value: *nats_ips
- type: replace
  path: '/instance_groups/name=log_controller/jobs/name=route_registrar/properties/nats/machines'
  value: *nats_ips

- type: replace
  path: '/instance_groups/name=consul/jobs/name=consul_agent/properties/consul/agent/servers/lan'
  value: *consul_static_ips
- type: replace
  path: '/instance_groups/name=nats/jobs/name=consul_agent/properties/consul/agent/servers/lan'
  value: *consul_static_ips
- type: replace
  path: '/instance_groups/name=etcd/jobs/name=consul_agent/properties/consul/agent/servers/lan'
  value: *consul_static_ips
- type: replace
  path: '/instance_groups/name=mysql/jobs/name=consul_agent/properties/consul/agent/servers/lan'
  value: *consul_static_ips
- type: replace
  path: '/instance_groups/name=diego-bbs/jobs/name=consul_agent/properties/consul/agent/servers/lan'
  value: *consul_static_ips
- type: replace
  path: '/instance_groups/name=uaa/jobs/name=consul_agent/properties/consul/agent/servers/lan'
  value: *consul_static_ips
- type: replace
  path: '/instance_groups/name=diego-brain/jobs/name=consul_agent/properties/consul/agent/servers/lan'
  value: *consul_static_ips
- type: replace
  path: '/instance_groups/name=diego-cell/jobs/name=consul_agent/properties/consul/agent/servers/lan'
  value: *consul_static_ips
- type: replace
  path: '/instance_groups/name=router/jobs/name=consul_agent/properties/consul/agent/servers/lan'
  value: *consul_static_ips
- type: replace
  path: '/instance_groups/name=tcp-router/jobs/name=consul_agent/properties/consul/agent/servers/lan'
  value: *consul_static_ips
- type: replace
  path: '/instance_groups/name=route-emitter/jobs/name=consul_agent/properties/consul/agent/servers/lan'
  value: *consul_static_ips
- type: replace
  path: '/instance_groups/name=blobstore/jobs/name=consul_agent/properties/consul/agent/servers/lan'
  value: *consul_static_ips
- type: replace
  path: '/instance_groups/name=api/jobs/name=consul_agent/properties/consul/agent/servers/lan'
  value: *consul_static_ips
- type: replace
  path: '/instance_groups/name=cc_clock/jobs/name=consul_agent/properties/consul/agent/servers/lan'
  value: *consul_static_ips
- type: replace
  path: '/instance_groups/name=cc_bridge/jobs/name=consul_agent/properties/consul/agent/servers/lan'
  value: *consul_static_ips
- type: replace
  path: '/instance_groups/name=doppler/jobs/name=consul_agent/properties/consul/agent/servers/lan'
  value: *consul_static_ips
- type: replace
  path: '/instance_groups/name=log_controller/jobs/name=consul_agent/properties/consul/agent/servers/lan'
  value: *consul_static_ips

- type: replace
  path: '/instance_groups/name=mysql/jobs/name=mysql/properties/cf_mysql/mysql/cluster_ips'
  value:
  - *mysql_static_ip
- type: replace
  path: '/instance_groups/name=uaa/jobs/name=uaa/properties/uaadb/address'
  value: *mysql_static_ip
- type: replace
  path: '/instance_groups/name=api/jobs/name=cloud_controller_ng/properties/ccdb/address'
  value: *mysql_static_ip
- type: replace
  path: '/instance_groups/name=api/jobs/name=cloud_controller_worker/properties/ccdb/address'
  value: *mysql_static_ip
- type: replace
  path: '/instance_groups/name=cc_clock/jobs/name=cloud_controller_clock/properties/ccdb/address'
  value: *mysql_static_ip
```

It's a bit long, but by using yaml anchors, you only have to specify your override IPs once at the top of the file. Would this solve your use case?

For the record, my only aversion to the change provided in this PR is that users would be required to provide two `-l` flags -- one for the default static ips, and another for the credentials.

rosenhouse:
@dsabeti Does this approach compose with using opts to add extra stuff to cf-deployment?  e.g. https://gist.github.com/cppforlife/cafa7d9149695d31e9b294195204bde3

dsabeti:
@rosenhouse I just tried the following:
```
bosh build-manifest -l deployment-env-vars.yml cf-deployment.yml -o netman.yml -o static-ip-overrides.yml
```

And got a manifest that looks has both the static-ip overrides and the netman additions. So I think this should be good.

@nelsam, any thoughts on overrides versus variables?

rosenhouse:
That's good to see.  

I still have a slight preference for variables for these things, with this repo providing the vars file for the bbl defaults.

Some reasons, in no particular order:
1) IPs are similar to `system_domain`, and I think they should be treated similarly

2) the ops file feels heavy-weight.  it is meant for overriding parts of a manifest that the original manifest author didn't anticipate would change.  But IPs are obviously something that will vary per-deployment.  Even bosh-lite will need custom IPs.

Just my $0.02

nelsam:
The ops file is very heavy-handed, but if all static IPs are going to be gone soon (replaced with bosh links), it won't be around for long.  If that's the case, then the ops file would end up being more friendly to teams that don't need to replace the IPs (i.e. they won't need to do anything when the last IPs are removed), and teams that are using it will actually get an error stating that the `static_ips` value no longer exists, prompting them to remove it.

On the other hand, the loggregator team is having a hard time making use of bosh links - our jobs can't `consume` bosh links without causing old `cf-release` deploys to fail because both `doppler_z1` and `doppler_z2` declare that they `provide` links named `doppler`.  If this is a blocker and we all need to be on `cf-deployment` before we can start consuming links, then the ops file is going to be around for a long time, and that makes me nervous.  My hope is that we can get around this with conditionals of some sort, but we haven't been able to find a conditional that works - bosh errors pretty early on when our jobs declare that they consume values.

rosenhouse:
@nelsam I think @Amit-PivotalLabs and @dsabeti  have some workarounds for the two-zones issue, something about consuming a `nil` link.

dsabeti:
@nelsam. So, in your cf-release deploys, you'll essentially need to opt out of using links. I think that's ok, since we're planning to get everyone over to cf-deployment over the next few months.

Like Gabe said, there's a way to do this. Typically, we use "implicit linking," which requires that both the provider and consumer of the link know it by the same name. Here's an example of a manifest that ***doesn't*** use implicit linking:
```
instance_groups:
- name: database
  jobs:
  - name: postgresql
    release: postgresql-release
    provides:
      db: { as: my-db }
- name: app
  jobs:
  - name: web
    release: my-app
    consumes:
      db: { from: my-db }
```
As you can see, the manifest explicitly renames the link to `my-db` in both the provider and consumer configuration.

We *could* decide to remove the re-naming of the link, since both the provider and the consumer call the link `db` by default:
```
instance_groups:
- name: database
  jobs:
  - name: postgresql
    release: postgresql-release
- name: app
  jobs:
  - name: web
    release: my-app
```
This is clearly much nicer, and we expect that we'll use this format most of the time.

So, in order to opt out of a link, we use the explicit consumpution of a link, but we `nil` it out:
```
instance_groups:
#- name: database
#  jobs:
#  - name: postgresql
#    release: postgresql-release
#    provides:
#      db: { as: my-db }
- name: app
  jobs:
  - name: web
    release: my-app
    consumes:
      db: nil
    properties:
      db_string: postgres://10.0.1.10:5432/mydb
```
In this case, we've explicitly dropped the link by specifying `consumes: { db: nil }`, and we use a property instead.

Let me know if that works for your cf-release deployments that include links.

dsabeti:
Ok, we've added the ops-file as of this commit. I'm going to close out this PR.

Thanks @nelsam for the original idea, and @rosenhouse for the additional feedback.

PR #19, opened 2016-11-02
Set routing_api.auth_disabled to false explicitly

nelsam:
This is the default behavior.  Adding it in just allows bosh ops to be run against it.

[#132676971]

Signed-off-by: Warren Fernandes <wfernandes@pivotal.io>


dsabeti:
@nelsam, the next story in our team's backlog is to get the routing-api set up properly: https://www.pivotaltracker.com/story/show/134618859. Since we're getting this to work by default, I'm going to close this PR.

PR #20, opened 2016-11-11
Add more granular bbs sql database connection parameters

jvshahid:
*Note*: Old `db_connection_string` is still supported for backward
 compatability but will later be removed

[#133715603]

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -283,7 +283,12 @@ instance_groups:
           - label: key-2016-06
             passphrase: "((diego_bbs_encryption_keys_passphrase))"
           sql:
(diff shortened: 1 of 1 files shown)


jaresty:
Hey @jvshahid!

Thanks for your pull request.  We have two issues that we'd like to address:

1. Please make the pull request against `develop` (not `master`)
2. Please remove the `db_connection_string` variable as part of this change.  If you do not, users will have to specify both styles of db connection string to avoid having the other "specified" using the raw bosh variable syntax (e.g. the string `"((db_connection_string))"` would be treated as a value alongside the intentionally specified host, port, etc if it was not explicitly set in the bosh var file.

Thanks!
Josh

jvshahid:
done

PR #21, opened 2016-11-13
Fix metron agent cert properties

mcwumbly:
The cert/key for metron agent moved and was causing deployments to fail, with the agent failing to start on the first etcd instance:

See: https://c2c.ci.cf-app.com/teams/main/pipelines/netman/jobs/trucker-test-upgrade/builds/211

```bash
cat /var/vcap/sys/log/metron_agent/metron_agent.stderr.log
...
panic: Could not initialize doppler connection pool: failed to load keypair: tls: failed to find any PEM data in key input
```

Moving these properties to match the most recent release of loggregator fixes the issue:

https://c2c.ci.cf-app.com/teams/main/pipelines/netman/jobs/trucker-test-upgrade/builds/213

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -8,9 +8,6 @@ properties_shared_across_every_instance_group:
       deployment: "((metron_agent_deployment_name))"
       protocols:
       - tls
-      tls:
-        client_cert: "((metron_metron_agent_tls_client_cert))"
-        client_key: "((metron_metron_agent_tls_client_key))"
       etcd:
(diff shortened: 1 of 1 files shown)


dsabeti:
Hi @mcwumbly, based on information from the loggregator team, we've eschewed TCP in favor of UDP for now. We have [a separate story](https://www.pivotaltracker.com/story/show/135643159) to use grpc instead.

As such, would you be willing to close this PR?

mcwumbly:
sure thing.

PR #23, opened 2016-11-15
Use bosh-cli var interpolation, remove many vars

rosenhouse:
BOSH CLI [v0.0.106](https://github.com/cloudfoundry/bosh-cli/releases/tag/v0.0.106) added support for partial interpolation:

> For example now `system_domain` variable in `https://api.((system_domain))` will be interpolated.

So we can remove many vars and replace them with interpolations of `system_domain`.

cc: @cppforlife

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -333,7 +333,7 @@ instance_groups:
           internal:
             hostnames:
             - uaa.service.cf.internal
-        url: "((uaa_url))"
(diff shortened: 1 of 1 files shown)


dsabeti:
Hey @rosenhouse, now that we have the var store that populates values in the same way, would you be ok closing this issue?

rosenhouse:
oh yeah, thanks

PR #24, opened 2016-11-17
use `variables` section for now to generate creds until releases are augmented with type information for properties

cppforlife:
sample usage after this PR:

```
$ bosh deploy cf-deployment.yml --vars-store env-repo/deployment-vars.yml -v system_domain=mydomain.com
```

havent tried deploying this yet.

require new bosh cli v0.0.114+.

cc @rosenhouse @Amit-PivotalLabs

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -5,30 +5,30 @@ properties_shared_across_every_instance_group:
     syslog_daemon_config:
       enable: false
     metron_agent:
-      deployment: "((metron_agent_deployment_name))"
+      deployment: "((system_domain))"
       tls:
-        client_cert: "((metron_metron_agent_tls_client_cert))"
-        client_key: "((metron_metron_agent_tls_client_key))"
+        client_cert: "((metron_agent_tls_client.certificate))"
+        client_key: "((metron_agent_tls_client.private_key))"
       etcd:
-        client_cert: "((etcd_client_cert))"
-        client_key: "((etcd_client_key))"
+        client_cert: "((etcd_client.certificate))"
+        client_key: "((etcd_client.private_key))"
     metron_endpoint:
       shared_secret: "((dropsonde_shared_secret))"
     loggregator:
       tls:
-        ca_cert: "((loggregator_tls_ca_cert))"
+        ca_cert: "((loggregator_ca.certificate))"
         trafficcontroller:
-          cert: "((loggregator_tls_tc_cert))"
-          key: "((loggregator_tls_tc_key))"
+          cert: "((loggregator_tls_tc.certificate))"
+          key: "((loggregator_tls_tc.private_key))"
         metron:
-          cert: "((metron_metron_agent_tls_client_cert))"
-          key: "((metron_metron_agent_tls_client_key))"
+          cert: "((metron_agent_tls_client.certificate))"
+          key: "((metron_agent_tls_client.private_key))"
         doppler:
-          cert: "((loggregator_tls_doppler_cert))"
-          key: "((loggregator_tls_doppler_key))"
+          cert: "((loggregator_tls_doppler.certificate))"
+          key: "((loggregator_tls_doppler.private_key))"
       etcd:
         require_ssl: true
-        ca_cert: "((etcd_ca_cert))"
+        ca_cert: "((etcd_server.ca))"
         machines:
         - etcd.service.cf.internal
 update:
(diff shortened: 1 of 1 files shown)


rosenhouse:
awesome.  it's so much better to have this built in to bosh.

cf-deployment.yml
```diff
@@ -5,30 +5,30 @@ properties_shared_across_every_instance_group:
     syslog_daemon_config:
       enable: false
     metron_agent:
-      deployment: "((metron_agent_deployment_name))"
+      deployment: "((system_domain))"
```
JesseTAlford: Why change this? This doesn't appear to be related to the new BOSH feature.

cf-deployment.yml
```diff
@@ -66,12 +66,12 @@ instance_groups:
           domain: cf.internal
           servers: &3
             lan: *1
-        encrypt_keys: "((consul_encrypt_keys))"
-        agent_cert: "((consul_agent_cert))"
-        agent_key: "((consul_agent_agent_key))"
-        ca_cert: "((consul_agent_ca_cert))"
-        server_cert: "((consul_agent_server_cert))"
-        server_key: "((consul_agent_server_key))"
+        encrypt_keys: [((consul_encrypt_key))]
```
JesseTAlford: the consul encrypt_keys are not necessarily a single value; an array needs to be provided for rotation. Why this change? Again, it doesn't appear to be related to the new bosh functionality.

cf-deployment.yml
```diff
@@ -152,12 +152,12 @@ instance_groups:
           servers: *3
           services:
             etcd: {}
-        encrypt_keys: "((consul_encrypt_keys))"
-        agent_cert: "((consul_agent_cert))"
-        agent_key: "((consul_agent_agent_key))"
-        ca_cert: "((consul_agent_ca_cert))"
-        server_cert: "((consul_agent_server_cert))"
-        server_key: "((consul_agent_server_key))"
+        encrypt_keys: [((consul_encrypt_key))]
```
JesseTAlford: Again, why the format switch?

cf-deployment.yml
```diff
@@ -168,24 +168,24 @@ instance_groups:
...
         require_ssl: true
-        ca_cert: "((etcd_ca_cert))"
-        client_cert: "((etcd_client_cert))"
-        client_key: "((etcd_client_key))"
-        server_cert: "((etcd_server_cert))"
-        server_key: "((etcd_server_key))"
-        peer_ca_cert: "((etcd_peer_ca_cert))"
-        peer_cert: "((etcd_peer_cert))"
-        peer_key: "((etcd_peer_key))"
+        ca_cert: "((etcd_client.ca))"
```
JesseTAlford: Shouldn't this just be `((etcd.ca))`? It's used to sign both the client and the server certs, isn't it?

cf-deployment.yml
cppforlife: not sure why that got replaced.

Amit-PivotalLabs:
Also this should be PR'd against `develop` @JesseTAlford @cppforlife @dsabeti

cf-deployment.yml
Amit-PivotalLabs: Yeah, it's awkward to use `"((etcd_client.ca))"` here, since it's the CA commonly used for both `etcd_client` and `etcd_server`.  More appropriate would probably be `"((etcd_ca.cert))"`.

cf-deployment.yml
Amit-PivotalLabs: The above is an array consisting of a single value.  The [`cf-filler` recipe](https://github.com/cloudfoundry/cf-deployment/blob/master/cf-filler/recipe-cf-deployment.yml) similarly only provides a password array with a single value.  It seems like right now, either you don't leverage something for automatic credential generation so that you can easily manage credential rotation yourself in your `-l` file, making the non-rotation and rotation deploys have similar UX, or you leverage automatic generation making the basic deploy easy but have to think about it a fair bit more when trying to do rotation.

Long run, it might be better for Config Server to support a password list type (instead of just a password type), but for this to be useful a lot of upstream code will need to change to respect password rotation via multiple passwords.

Amit-PivotalLabs:
If this were to be merged, the README should also be updated and `cf-filler` directory removed, as it would be confusing to have two ways to do this.

cf-deployment.yml
rosenhouse: That's how it is in the [cf-filler recipe](https://github.com/cloudfoundry/cf-deployment/blob/master/cf-filler/recipe-cf-deployment.yml#L18).  I set that originally because it seemed like a sensible default, and was one less decision that the user had to make.

jaresty:
Can't merge this yet since consul agent/server require Extended Key Usage enabled for authentication, but bosh-cli doesn't have a way to enable it yet. @cppforlife

Amit-PivotalLabs:
@jaresty did you try this, and it failed?  Were there error messages in the consul logs when trying this?

@cppforlife @tylerschultz @pivotal-danjahner @ishustava config server/CredHub may need to support [Extended Key Usage](https://www.consul.io/docs/agent/encryption.html#rpc-encryption-with-tls) enablement for cert generation to be viable for usage by Cloud Foundry deployments.

jaresty:
Yes. There were some issues with the certificates when I first tried to
deploy it. I've since been able to get it to deploy successfully and consul
dropped fine with a single instance on bosh lite using this fork of the cli
https://github.com/jaresty/bosh-cli/tree/ext-key-usage with the var-store
branch of cf-deployment (which is slightly modified from the initial pr)
https://github.com/cloudfoundry/cf-deployment/tree/var-store?files=1
2016年11月30日(水) 17:21 Amit Gupta <notifications@github.com>:

cppforlife:
@jaresty @dsabeti can we close this pr at this point?

jaresty:
@dsabeti Do you want to accept it before you close the PR?  I think we are
good; just waiting for it  to pass CI now.

dsabeti:
This functionality has been merged in commit 07c1da1, so I'm going to close this PR. Thanks @cppforlife!

PR #26, opened 2016-11-22
Add a windows cell

dgodd:
Default instances for windows cells is set to 0 since it is assumed that most people won't want any

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -542,6 +542,48 @@ instance_groups:
   - name: metron_agent
     release: loggregator
     properties: *2
+- name: windows-cell
+  azs:
+  - z1
+  - z2
+  instances: 0
+  vm_type: r3.xlarge
+  vm_extensions:
+  - 100GB_ephemeral_disk
+  stemcell: windows
+  networks:
+  - name: private
+  jobs:
+  - name: consul_agent_windows
+    release: consul
+    properties:
+      consul:
+        agent:
+          mode: client
+          domain: cf.internal
+          servers: *3
+        encrypt_keys: "((consul_encrypt_keys))"
+        agent_cert: "((consul_agent_cert))"
+        agent_key: "((consul_agent_agent_key))"
+        ca_cert: "((consul_agent_ca_cert))"
+        server_cert: "((consul_agent_server_cert))"
+        server_key: "((consul_agent_server_key))"
+  - name: garden-windows
+    release: garden-windows
+  - name: rep_windows
+    release: diego
+    properties:
+      diego:
+        rep:
+          bbs: *5
+          require_tls: false
+          preloaded_rootfses:
+          - windows2012R2:/tmp/windows2012R2
+        ssl:
+          skip_cert_verify: true
+  - name: metron_agent_windows
+    release: loggregator
+    properties: *2
 - name: router
   azs:
   - z1
@@ -862,6 +904,11 @@ instance_groups:
         resource_pool: *7
         packages: *7
         droplets: *7
+        stacks:
+        - name: "cflinuxfs2"
+          description: "Cloud Foundry Linux-based filesystem"
+        - name: "windows2012R2"
+          description: "Windows Server 2012 R2"
       ccdb: &13
         databases:
         - name: cloud_controller
@@ -1368,7 +1415,14 @@ releases:
   url: https://bosh.io/d/github.com/cloudfoundry/staticfile-buildpack-release?v=1.3.13
   version: 1.3.13
   sha1: 0741ad6d39cf258c317151e1227e0ccfe1fd7579
+- name: garden-windows
+  version: 0.0.9
+  url: https://bosh.io/d/github.com/cloudfoundry-incubator/garden-windows-bosh-release?v=0.0.9
+  sha1: c72b842e9b18c6fc2525f3ac9ec3973fb3797e47
 stemcells:
 - alias: default
   os: ubuntu-trusty
   version: '3312'
+- alias: windows
+  os: windows2012R2
+  version: latest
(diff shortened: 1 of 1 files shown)


cppforlife:
@JesseTAlford is this better server via an opsfile?

cppforlife:
also windows stemcell is missing.

mavenraven:
@cppforlife what do you mean by it's missing? looks like the manifest doesn't pull the trusty one from bosh.io either

dgodd:
@cppforlife This could possibly be accomplished with an opsfile, if opsfiles did additions (currently just replace / remove). However we would like to make it easy for internal teams to deploy windows stemcells, so that it is easy for them to discover if they are about to break windows.

I would also to add to David's question about stemcells, we have the added problem that the latest stemcell on windows has a different version per IAAS since we can't upload them or distribute them due to licensing issues (unlike linux).

cppforlife:
inline

Sent from my iPhone

> On Nov 30, 2016, at 7:15 AM, Dave Goddard <notifications@github.com> wrote:
> 
> @cppforlife This could possibly be accomplished with an opsfile, if opsfiles did additions (currently just replace / remove).
> 
replace allows additions. see https://github.com/cppforlife/go-patch/blob/master/docs/examples.md
> However we would like to make it easy for internal teams to deploy windows stemcells, so that it is easy for them to discover if they are about to break windows.
> 
> I would also to add to David's question about stemcells,
> 
i didnt see update to stemcells section first time. thats what i meant.
> we have the added problem that the latest stemcell on windows has a different version per IAAS since we can't upload them or distribute them due to licensing issues (unlike linux).
> 
topic for a diff thread but versions can be synced up. it s not like its entirely diff windows distro.

JesseTAlford:
I agree that this could be better accomplished with an opsfile, until/unless running with a Windows stemcell is our default configuration.

dgodd:
@cppforlife Thank you for pointing out additions can be done using the `go-patch` syntax. Unfortunately I did not realize I should be looking at `go-patch` for the syntax. 

My next problem with using an opsfile is that the `diego-cell` which is what we need copied, currently using multiple references (eg to consul servers etc...) I tried including the references (eg. *2) in the file but this does not work (my assumption but I thought I should check just in case).

If there was a way to duplicate/copy a node and then alter it afterwards that might work?

As a side note, I would like to believe that many more teams will start running windows cells in the near future to make sure that they don't change things such that windows no longer works. (and I did set `instances: 0` for the windows cells)

cppforlife:
> If there was a way to duplicate/copy a node and then alter it afterwards that might work?

there isnt a way to dup/copy node.

Given that I guess instances: makes sense. Im not sure bosh would be happy though if you mention a stemcell in stemcells section that isnt uploaded to the Director.

dgodd:
I just tried two options

1. A Job with a missing stemcell
2. A stemcell with a verison which is not uploaded

Neither works

The only option I can think of is to set the windows-cell to use the **default** stemcell, and change it to windows (and add the windows stemcell) in an ops file.

Or add **dup** to go-patch ; would you be interested in my submitting that PR?

JesseTAlford:
Or, fully specify windows cell configuration alongside the standard
configuration in an ops file.

Philosophically, cf-deployment.yml is meant to contain nothing except what
is necessary for a full-featured standard Cloud Foundry deployment.
Anything else lives in ops files. If future bosh features make those ops
files simpler, fine. If they don't, optional properties still don't belong
in cf-deployment.yml. Duplication is preferable.

As releases publish and consume bosh links, we expect most duplication to
evaporate anyway.

dgodd:
@JesseTAlford My understanding of opsfiles is that they should be fairly generic, is that true? In which case the windows one is very problematic, since adding or subtracting consul instances or anything else will immediately break the windows opsfile.

I think there is a whole seperate question around **full-featured standard Cloud Foundry deployment**, should windows be part of that? My belief is that the we (cloud foundry) are pushing towards this.

mavenraven:
@JesseTAlford What's the problem we're trying to solve by not including it by default? We absolutely should be considering Windows cells as a first class citizen on the platform.

dsabeti:
Hey @dgodd && @mavenraven.  Let me explain a bit here.

When we say that cf-deployment is a "full-featured standard Cloud Foundry deployment," we mean the following things:
1. cf-deployment.yml can be deployed without any changes (except for the required `-l vars.yml` for environment-specific information).
1. cf-deployment uses default configuration wherever possible. Any customization should be done with ops files.
1. In the same way, the selection of instance groups and releases should represent a "default" deployment of CF. Undoubtedly, this represents a certain opinion about how to deploy CF. Any deviations from this default should be considered extensions -- extensions which may still add in "first class" or fully supported features, just not features that are used by default.
1. Excluding extensions, cf-deployment opts in to all features of CF.

With regards to "extensions," I'd say that being an extension does not mean that it doesn't represent a "first class" feature. For example, there's an ops file for deploying to GCP, even though the platform is considered fully supported. We're planning to add ops files for using postgresql as your DB, which is also a first class feature. We want to avoid polluting cf-deployment with a bunch of superfluous releases and instance groups, even if they're configured to 0 instances by default. We'd rather allow for ops files to bring in those releases and instance groups only if the user is intending to deploy with that configuration. 

So, in the case of windows-cells, we'd support a `windows.yml` that updates the manifest such that it can deploy the windows cell with a windows stemcell. If we merged the change into cf-deployment.yml directly, every user of cf-deployment would have to upload a windows stemcell and garden-windows release to their director, even if they had no intention of deploying the windows cell. With ops files, you can choose to deploy one or the other (or both, I suppose). To show that windows would still be a first class feature, we'd save the `windows.yml` in the `opsfiles` directory of cf-deployment, and we could also test it in CI before cutting releases.

Does that answer your questions?

dsabeti:
@dgodd @mavenraven After our conversation this week, I've written a story to get this working as an ops file: https://www.pivotaltracker.com/story/show/135641715. Can I close out this PR?

aminjam:
Yes. Let us know when we should x-teampair on that.

PR #31, opened 2016-12-19
Bosh lite fixups

rosenhouse:
Without these changes, I couldn't get the bosh-lite deploy to start.

Relatedly: It'd be awesome if you were running bosh-lite in your CI pipeline!


JesseTAlford:
Thanks for the PR, @rosenhouse! We're actually doing the `cf-deployment`-on-`bosh-lite`-in-CI thing right now! I've stuck this at the top, we should pull it in shortly, maybe even as part of that work.

PR #35, opened 2016-12-23
Set router_ssl alternative names

acrmp:
We noticed recently that the router_ssl certificate alternative_names were not being set.

Also taking the opportunity to add `*.((system_domain))` so that the certificate is valid for subdomains.

Thanks,

Andrew && Stu


JesseTAlford:
Thanks! These are both good changes.

PR #39, opened 2017-01-05
use `ports` VM cloud property on bosh-lite

cppforlife:
- added vm cloud properties to bosh-lite cloud config
- renamed tcp-routing to not have gcp in the name (nothing is iaas specific in this repo!)

[future] need to add ops file for adding external ip (bosh-lite' host vm ip) to gorouter_ssl certificate so that cf cli ca properly use --ca-cert and verify certs


jaresty:
@cppforlife Thanks for submitting this PR!  We are excited to get port forwarding working so we can have bosh-lite deployed successfully to GCP.  I have a couple of questions about this PR and I figured I should ask right away since we want to get moving on the bosh-lite story as soon as possible:

1. Is  the `latest-stemcell.yml` ops file required for this PR to work?
2. Can we treat the GCP TCP router ops-file rename as a separate issue?  If so, do you mind if we revert that change for now?
  * I did notice that you added a `cf-tcp-router-network-properties` vm_extension to the cloud-config but I expect that if nothing is tagged with the `cf-tcp-router-network-properties` then it will just be ignored -- right?
3. With respect to the `[future] need to add ops file for adding external ip` comment in your PR: If the bosh director is available via an asterisk DNS rule at `api.snitch.cf-app.com` as it will be for this GCP bosh-lite director, then I would expect that the SSL certificate should work via DNS and so the IP address difference would not matter.  Is my assumption incorrect?

PR #41, opened 2017-01-13
Require mutual TLS for Windows consul agent API

mhoran:
This change requires TLS between the rep and the consul agent on Windows. Note that this can't be merged until https://github.com/cloudfoundry/diego-release/pull/253 is merged, but we're sending it now so that it can be merged when everything is in place.

### DIFF
--- a/operations/windows-cell.yml
+++ b/operations/windows-cell.yml
@@ -21,6 +21,7 @@
           enable: false
         consul:
           agent:
+            require_ssl: true
             mode: client
             domain: cf.internal
             servers:
@@ -51,6 +52,11 @@
               ca_cert: "((diego_bbs_client.ca))"
               client_cert: "((diego_bbs_client.certificate))"
               client_key: "((diego_bbs_client.private_key))"
+            consul:
+              ca_cert: "((consul_agent.ca))"
+              client_cert: "((diego_consul_client.certificate))"
+              client_key: "((diego_consul_client.private_key))"
+              require_tls: true
             require_tls: false
             preloaded_rootfses:
             - windows2012R2:/tmp/windows2012R2
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -1386,6 +1386,8 @@ variables:
     ext_key_usage:
     - client_auth
     - server_auth
+    alternative_names:
+    - 127.0.0.1
 - name: consul_server
   type: certificate
   options:
@@ -1428,6 +1430,13 @@ variables:
     alternative_names:
     - "*.bbs.service.cf.internal"
     - bbs.service.cf.internal
+- name: diego_consul_client
+  type: certificate
+  options:
+    ca: consul_agent_ca
+    common_name: diego consul client
+    ext_key_usage:
+    - client_auth
 - name: loggregator_ca
   type: certificate
   options:


aminjam:
@dsabeti This PR is ready to be pulled in. We now have [Diego 1.6.0](https://github.com/cloudfoundry/cf-deployment/commit/444c97f7fb239a85a0f4d7d9ff90bb0c870bc242) that has our mutual TLS changes.

sunjayBhatia:
@dsabeti any word on pulling this in? we're trying to ensure all of our environments and our consumers have this security setting properly set

mhoran:
Hey @jaresty. The new cert and key are used to secure communication between consumers of the consul HTTP(S) API and the consul agent. The existing cert and key are used by the agent itself for securing communication to the consul server. While we could have used that same certificate, that would preclude us from ever revoking the Diego certificate independently of the agent certificate. We also felt that this was an odd way to reuse a certificate that is already used for other purposes.

Also, we could generate a unique CA for the HTTPS API and load that into the agent only, protecting the greater consul cluster from any bad actor who manages to extract the Diego certificate and key. For simplicity, we decided not to do so at this time.

jaresty:
Hi @mhoran,

Thanks for getting back to us.  We still don't understand completely so would like to discuss on audio before moving ahead.  Please let us know the best way to reach you.  You have a message on pivotal slack from us.  

Thanks,
Josh && @mingxiao

jaresty:
@mhoran I think we are starting to get your drift, but would to confirm: when you say
> new cert and key are used to secure communication between consumers of the consul HTTP(S) API and the consul agent
I think you mean that the consumer (in this case, Diego, or some part of Diego) is talking to the consul server directly via https.  Does that sound correct?  Thanks!

sunjayBhatia:
@jaresty The rep has a consul client and is talking to the consul agent on the cell via https

jaresty:
@sunjayBhatia Ok.  I think I understand now.

operations/windows-cell.yml
```diff
@@ -51,6 +52,11 @@
               ca_cert: "((diego_bbs_client.ca))"
               client_cert: "((diego_bbs_client.certificate))"
               client_key: "((diego_bbs_client.private_key))"
+            consul:
+              ca_cert: "((consul_agent.ca))"
```
jaresty: Please change this to `diego_consul_client.ca` to match the current pattern where we reference the CA via the name of the certificate directly rather than configuring a CA that we "know" to be the same.

mhoran: Hey @jaresty. This is a bit confusing, but the `ca_cert` here is used by the rep to determine if it's talking to a trusted agent. The agent is running with the `consul_agent.ca`, not the `diego_consul_client.ca` (which do happen to be the same in this case.)

The rep then presents its certificate to the agent for mutual TLS. The agent will verify that the `diego_consul_client.certificate` was signed by the `diego_consul_client.ca`, which happens to be the same as `consul_agent.ca`. So if you really wanted us to, we would change [this line](https://github.com/cloudfoundry/cf-deployment/pull/41/files/be40a85443c4e6f76bf718eb5725f4145eb3a23a#diff-f73030c6ed757d99f2745c7d5d85d186R36) to include both the `diego_consul_client.ca` and `consul_agent.ca`, but they'd be identical and I don't know what that would do.

jaresty: @mhoran I think I understand.  In that case, is there some way that we can make this difference more obvious?  I think that calling it `ca_cert` (which looks so much like our other properties where the ca_cert is used to sign the certificate which is located adjacent in the manifest) makes it especially easy to misunderstand in this case.  @dsabeti && @anEXPer , do you want to weigh in here?

Thanks,
Josh && @mingxiao

dsabeti: @jaresty Your suggestion about renaming `ca_cert` would require changes to the release -- not something that makes sense to do.

As far as using `consul_agent.ca` vs `diego_consul_client.ca`, I think it's fine to keep as `consul_agent.ca` for now.

Let's go ahead an merge this is in now.

PR #44, opened 2017-01-18
Add opsfiles for local-route-emitter mode

anoop2811:
- There is an opsfiles for enabling local route emitters on diego cells.
- There is an opsfiles for disabling global route emitters.

cc @genevievelesperance


anoop2811:
(Gen) After speaking with RelInt, we are going to add some duplication to one of the opsfiles to remove the need for explaining "opsfile dependencies" for the local-route-emitters mode feature.

PR #45, opened 2017-01-18
Add opsfiles for local-route-emitter mode

anoop2811:
- There is an opsfile for enabling local route emitters on diego cells
*and* removing global route emitters.
- There is an opsfile for enabling local route emitters on diego cells.

There is duplicate code between these opsfiles. We debated some sort of assertion logic that would validate that the local route emitters were enabled if global route emitters were to be disabled. This seemed rather brittle. After discussion with RelInt, we figured the duplicate code would eventually be refactored in favor of some sort of assertion in the opsfile.

cc @genevievelesperance


anoop2811:
We've updated this PR to include opsfiles for enabling local-route-emitter mode on windows cells.

PR #52, opened 2017-01-30
Added is_ca generation parameter to CAs

pivotal-danjahner:
Also moved the leaf certificates of 'service_cf_internal_ca' under the CA for better readability


pivotal-danjahner:
Will simplify to just 'is_ca' addition to avoid conflicts

PR #60, opened 2017-02-03
Configure syslog drain binder-CC mutual TLS

jtarchie:
We'd like these add these BOSH properties.

It requires that this [story](https://www.pivotaltracker.com/story/show/138835189) be accepted before it is merged. This way we don't introduced unused properties before they are consumed.


enocom:
We are merging this as it shouldn't affect anyone.

PR #61, opened 2017-02-08
Update statsd-injector

bradylove:
We have updated the statsd-injector to use Loggregator V2 API. 

@enocom


anEXPer:
Looks good to me. I'm also re-enabling automatic bumps of the statsd-injector release; we had paused until we had working generation for these properties.

Thanks!

PR #62, opened 2017-02-08
Allow consul to consume its own link

wendorf:
Consul can be run in two modes, server or client. We only want the
server-mode consul jobs to provide a link, but due to the nature of bosh
links, all consul jobs provide the link. Therefore, we need to select
the correct link.

Note: This PR should not be merged until cf-deployment is consuming a consul-release that includes cloudfoundry-incubator/consul-release#64, which will presumably be consul-release v153.

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -21,20 +21,18 @@ instance_groups:
     serial: true
   networks:
   - name: private
-    static_ips: &consul_ips
-    - 10.0.31.190
-    - 10.0.47.190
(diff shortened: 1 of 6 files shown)


wendorf:
consul-release v154 does not use the IP addresses in the link, so this PR will not work. This PR should only be merged once https://github.com/cloudfoundry-incubator/consul-release/commit/22515a9805d878022951d4654eac6dbeafc7fe64 is released (presumably in consul-release v155)

PR #63, opened 2017-02-08
Remove nats_stream_forwarder

drich10:
We are working under the assumption that it is not necessary in the
current cf-deployment universe where nats is used much less. If this
turns out not to be the case, we can bring it back.

Note: nats_stream_forwarder does not currently consume the nats link,
so if it is brought back, the release should be updated first.

Signed-off-by: Derek Richard <drichard@pivotal.io>

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -95,14 +95,6 @@ instance_groups:
         ca_cert: "((consul_agent.ca))"
         server_cert: "((consul_server.certificate))"
         server_key: "((consul_server.private_key))"
-  - name: nats_stream_forwarder
-    release: nats
-    properties:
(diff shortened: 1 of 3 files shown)


jaresty:
Hi guys,

I made the same change [over here](https://github.com/cloudfoundry/cf-deployment/commit/a8f17c09e78bb5d2e23a5b313da6e19046cf45cb) since it was easier for me to be certain I didn't make a mistake with the merge conflict.  I tried to git duet you both but made a mistake; so, sorry -- looks like drich10 committed with pivotal.  I hope this resolves your issue.  We will watch to make sure it goes through CI.

drich10:
All good! Thanks for merging :)

PR #71, opened 2017-02-17
Add override file to swap mysql for postgres

ljfranklin:
- WARNING: We also changed variable names for the cc, diego, and uaa
  databases from `cf_mysql_mysql_seeded_databases_cc_password` to
  `cc_database_password`. This will cause bosh to roll that secret and
  the next deploy might be a bit bumpy.

[#139907835]

Signed-off-by: Lyle Franklin <lfranklin@pivotal.io>

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -223,13 +223,13 @@ instance_groups:
           seeded_databases:
           - name: cloud_controller
             username: cloud_controller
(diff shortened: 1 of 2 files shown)


ljfranklin:
We're new to writing ops-files so any stylistic feedback is welcome!

PR #77, opened 2017-02-22
Remove nats config from cloud_controller jobs

wendorf:
As of capi-release 0.121.0, cloud_controller jobs can consume nats links.


mingxiao:
We did a couple little clean-up commits right after the merge to stop referencing these properties in the bosh-lite ops file.

Thanks for the PR!

PR #79, opened 2017-02-23
route_emitter does not use nats config in operations

wendorf:
Use link instead

[#139909121]

This goes along with #78


ematpl:
Thanks, @wendorf! You probably want to clear the NATS config out of https://github.com/cloudfoundry/cf-deployment/blob/master/operations/enable-local-route-emitters-windows-cell.yml and https://github.com/cloudfoundry/cf-deployment/blob/master/operations/enable-local-route-emitters.yml as well.

Best,
Eric

wendorf:
Good catch! I was foiled by underscores vs dashes. Fixed.

PR #83, opened 2017-02-28
Add mutual_tls properties to the CC clock

aashah:
This property is not exercised by the currently configured capi-release (1.21.0), but there's no harm in adding it. This property will be necessary for the cc-based bridge components.

[#139564797]

Signed-off-by: Eric Promislow <eric.promislow@gmail.com>


aashah:
This work seems to have already been as part of c1d9ca78d25ec3a3159e35ca5a852c37488de9c3. I'm going to close this issue.

PR #87, opened 2017-03-06
Update routers in serial

shashwathi:
This PR will enable routers to update serially and not in parallel. 

Regards
Shash

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -232,6 +232,8 @@ instance_groups:
   - z1
   - z2
   instances: 2
+  update:
+    serial: true
   vm_type: m3.medium
   vm_extensions:
   - router-lb


cppforlife:
out of curiosity why make them serial?

anEXPer:
This is a great change. Thanks! How did you come to notice it, out of curiosity?

anEXPer:
@cppforlife, how else would you suggest avoiding app downtime while rolling the routers?

cppforlife:
as long as max_in_flight is configured appropriately you are fine. moreover, max in flight is also scoped to within an az so currently routers will be updated one at time even though max in flight is 5.

anEXPer:
Maybe I'm not understanding, but if max-in-flight is per-AZ, doesn't that mean that each AZ can be rolling up to 1 instance at a time, so if each AZ only _has_ one instance, max-in-flight doesn't keep them from going at the same time at all?

anEXPer:
We found this created unexpected issues, because router needs UAA to come up, and it wasn't.

We didn't think through what the `serial: true` property actually does. If we really want this, we need to re-order how the jobs deploy. We might want to do that anyway.

shashwathi:
@anEXPer : 

If routing api is enabled, `gorouter` tries to connect to UAA to obtain token to connect to API server and this required UAA to be up and running before gorouter.
We probably did not encounter this before because UAA came up along with gorouter.

> If we really want this, we need to re-order how the jobs deploy. 

I agree with that change. Do you need PR from the routing team or XTP with us?  We are currently blocked on this change so we would get this change in cf-deployment ASAP.

Regards
Shash

PR #92, opened 2017-03-11
Add cfdot ops-file

aashah:
There was a [story](https://www.pivotaltracker.com/story/show/141530239) in Diego's icebox, but we had a need for this so we created an ops-file for it.

We chose not to add this to the cc-bridge, let us know if you would like that to be in there


ChunyiLyu:
Thanks for contributing to cf-deployment @aashah! After we merged the PR, it broke our deploy jobs because we already have a cfdot job on diego-cell instance_groups. In addition, our team tries to avoid re-specification of values that are the same as default values we see in job spec files. This PR specifies cfdot.bbs.use_ssl to be true, which is the default value, so we removed that line. [Here is the link to our style guide for contributing to cf-deployment](https://github.com/cloudfoundry/cf-deployment#editorial-style-guide).  We already pushed the fix, however, in the future, please follow our style guide. 

Thanks,
Release Integration Team

PR #94, opened 2017-03-15
Use port 80 for TCP health check

shashwathi:
Hi, 

Currently TCP health check port is set to ephemeral port( `8080`) by default but it is recommended for operators to use reserved port(`80`). If port `8080` is used for health check it could potentially be a clash with router groups. 

Story: https://www.pivotaltracker.com/story/show/141653531

Regards
Shash

### DIFF
--- a/operations/tcp-routing-gcp.yml
+++ b/operations/tcp-routing-gcp.yml
@@ -24,7 +24,7 @@
       properties:
         tcp_router:
           oauth_secret: "((uaa_clients_tcp_router_secret))"
-          health_check_port: 8080
+          health_check_port: 80
         uaa:
           ca_cert: "((uaa_ca.certificate))"
           tls_port: 8443


dsabeti:
@shashwathi We've wanted to do this change for a while (tracker story is [here](https://www.pivotaltracker.com/story/show/139259373)), but we're waiting for some necessary updates to `bbl`. 

Right now, `bbl`'s firewall rules prevent traffic from reaching port 8080. As a result, we have to stick with `bbl` 2.1.1 (which uses 8080 for healthchecking) until a new release uses 8080 for health checks and sets up firewall rules to allow that traffic. I'll let you know when we have the right version of `bbl` and can merge in this PR.

cc @evanfarrar

shashwathi:
@dsabeti : If I am understanding this correctly , are you saying that you cannot allocate port `8080` to TCP router groups in cf-deployment ? 

If this is the case, we probably should consider adding a note in README about this gotcha. 

Regards
Shash

dsabeti:
@shashwathi Yeah, that's temporarily the case until we get the necessary changes to `bbl` and start using the default port for healthchecking. I'll find a place to add that in the README.

PR #95, opened 2017-03-15
Disable legacy unsecured endpoints on the rep API

ematpl:
This change configures the rep to serve only its administrative
endpoints (/ping, /evacuate) over plain HTTP. This change also
configures this plain-HTTP server to listen only on localhost
and not a remotely routable IP.

[#140322915]


anEXPer:
Hey @ematpl,

Why are insecure endpoints enabled by default? We try and avoid unnecessary properties in cf-d; this seems like a good candidate for being pushed down to the spec.

Assuming disabling the insecure endpoints can't be made the default now, is there a plan for changing it to be the default in the future?

ematpl:
Backwards compatibility, will change in Diego v2.

PR #97, opened 2017-03-21
Fix clients that failed uaa manifest validation

bsekar:
Update uaa configuration to pass new configuration validations

These changes follow those previously submitted for cf-release 
https://github.com/cf-identity/cf-release/commit/de6f6f0612dfe8a77333bc422167e6dc25db166c

[#141162355] https://www.pivotaltracker.com/story/show/141162355

Signed-off-by: bsekar <bharath.sekar@ge.com>

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -305,6 +305,7 @@ instance_groups:
             override: true
             refresh-token-validity: 2592000
             scope: cloud_controller.read,cloud_controller.write,openid,password.write,cloud_controller.admin,scim.read,scim.write,doppler.firehose,uaa.user,routing.router_groups.read,routing.router_groups.write
+            secret: ''
           cloud_controller_username_lookup:
             authorities: scim.userids
             authorized-grant-types: client_credentials
@@ -316,7 +317,7 @@ instance_groups:
             secret: "((uaa_clients_doppler_secret))"
           gorouter:
             authorities: routing.routes.read
-            authorized-grant-types: client_credentials,refresh_token
+            authorized-grant-types: client_credentials
             secret: "((uaa_clients_gorouter_secret))"
           ssh-proxy:
             authorized-grant-types: authorization_code
@@ -327,11 +328,11 @@ instance_groups:
             secret: "((uaa_clients_ssh-proxy_secret))"
           tcp_emitter:
             authorities: routing.routes.write,routing.routes.read,routing.router_groups.read
-            authorized-grant-types: client_credentials,refresh_token
+            authorized-grant-types: client_credentials
             secret: "((uaa_clients_tcp_emitter_secret))"
           tcp_router:
             authorities: routing.routes.read,routing.router_groups.read
-            authorized-grant-types: client_credentials,refresh_token
+            authorized-grant-types: client_credentials
             secret: "((uaa_clients_tcp_router_secret))"
       uaadb:
         address: sql-db.service.cf.internal


anEXPer:
Thanks for the pull request, @bsekar.

I'd appreciate some context on some of these changes. Scoping down the router client's authorized grant types makes fine sense to me. Why is it necessary to provide an empty value for the `cf` client?

We're generally at pains to avoid duplicating specification of any default behavior in cf-d.

cf-deployment.yml
```diff
@@ -305,6 +305,7 @@ instance_groups:
             override: true
             refresh-token-validity: 2592000
             scope: cloud_controller.read,cloud_controller.write,openid,password.write,cloud_controller.admin,scim.read,scim.write,doppler.firehose,uaa.user,routing.router_groups.read,routing.router_groups.write
+            secret: ''
```
dsabeti: I'm curious about this line. One of the goals of cf-deployment is to remove as much unneeded configuration from the manifest as possible. Any time we provide an empty string in the manifest, it seems like an opportunity to ask about default values provided by the release.

Can you explain why we need to explicitly set this to an empty string? What is the default value if you don't set this value?

bsekar:
Hi @anEXPer @dsabeti , as part of [this story](https://www.pivotaltracker.com/n/projects/997278/stories/141101089), UAA validates oauth clients in the manifest. `secret` is a required property for clients that have scopes other than `implicit`. The secret for the client `cf` happens to be an empty string and the same needs to be specified in the yml.

cf-deployment.yml
fhanik: Absolutely. We are moving away from defaults on passwords, private keys, etc. An empty string is still a deliberate value. So if a client is configured for this value, it has to be spelled out. 

This makes the manifest very clear on what values are required and what the actual value is. Even if operator decides to set the password of a client to `""`

dsabeti: @fhanik Makes sense not to have a default for password. Should `secret` have a real (i.e. securely and randomly generated) value instead of an empty string? What effect would that have?

fhanik: @dsabeti the `cf` client uses a password grant. This means the client has to have a password.
In most situations, clients that cannot protect its credentials use the `implicit` grant. But this is a browser flow, thus we have the situation where the `cf` client is a command line application, but cannot protect its credentials.

The compromise was to have the client use the `password` flow and hard code the password to an empty string `""`. 

If you change the secret of the `cf` client then I believe `cf cli` application will stop working, unless it has the option to set the client password.

dsabeti: Thanks for this context @fhanik. I've prioritize the story to merge this.

bsekar:
@dsabeti @ChunyiLyu One of the changes in this pull request has not made it to the develop branch as part of the [merge](https://github.com/cloudfoundry/cf-deployment/commit/ffd8e9f54960502172cc0d6afc832f1cdafd78f9). The missing change is [here](https://github.com/cloudfoundry/cf-deployment/pull/97/files#diff-f39617caec1965e56248ce1bf556db4dL319).

staylor14:
@besekar sorry about that!  The pull request had conflicts, and when we resolved them, we overlooked an important change.

Fixed on develop @c995000

--Steve and @ChunyiLyu

bsekar:
@staylor14 cool, thanks!

PR #98, opened 2017-03-22
Update bosh-lite-on-gcp-deployment-guide.md

flawedmatrix:
The `bosh-deployment/gcp/cpi.yml` ops file will set the director's `cpi_job` property to `google_cpi` and overwrite the `cpi_job: warden_cpi` set by the `bosh-deployment/bosh-lite.yml` ops file.

When this happens, uploading a Bosh Lite Warden stemcell will yield an error:

`Error: CPI error 'Bosh::Clouds::CloudError' with message 'Creating stemcell: Invalid 'warden' infrastructure' in 'create_stemcell' CPI method`

Reversing the order of these two ops files fixes the problem.


dsabeti:
Great catch @flawedmatrix!

PR #100, opened 2017-03-23
Set default quota route ports to 100

mmb:
This matches the number of reservable ports in operations/tcp-routing-gcp.yml.
If this is not set the default is 0 and you need cf update-quota to allow
routes to be created with ports.


anEXPer:
Interesting. This raises the issue, for me, of why we are setting up a "default" quota. It seems like it might belong in the spec, but I can see why that might be hard or problematic...

PR #101, opened 2017-03-23
Set default app memory to 256M on bosh-lite

mmb:
This is what the cf-release bosh-lite manifest templates use.

### DIFF
--- a/operations/bosh-lite.yml
+++ b/operations/bosh-lite.yml
@@ -242,3 +242,8 @@
 - type: replace
   path: /instance_groups/name=log-api/instances
   value: 1
+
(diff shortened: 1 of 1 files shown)


dsabeti:
@mmb Just for full context, is there a specific need to reduce the default memory limit? For example, are you finding that tests fail as a result? Or that you can't push as many apps as you were able to on bosh-lite for cf-release?

mmb:
We can't run as many tests in parallel because they fail with insufficient resources errors.

We are already overriding this with our own ops file but we thought it made sense upstream because bosh-lite usually only has enough memory to push a few apps at 1G.

PR #102, opened 2017-03-23
Implement diego-deployment constraints

luan:
Currently `diego-bbs` and `diego-brain` are being deployed along with the cells, that can (and likely will) cause downtime.
This commit fixes that issue.

Source: https://github.com/cloudfoundry/diego-release/blob/develop/docs/deployment-constraints.md#diego-manifest-jobs

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -737,6 +737,8 @@ instance_groups:
   vm_extensions:
   - diego-ssh-proxy-network-properties
   stemcell: default
+  update:
+    serial: true
   networks:
   - name: private
   jobs:


anEXPer:
Thanks for the pull request!

Can you help me understand why _both_ of these need `serial: true` instead of just the brain?

luan:
@anEXPer per the docs:

> BBS servers should update before BBS clients.*...*
BBS clients include the brain and the cells, if BBS doesn't have serial true there's a risk that those will start before it, risking them to be hitting new APIs that don't yet exist.

> `brain_zN` jobs update separately from cells
When evacuating the cells the brain is instrumental in placing the evacuating instances into new cells, so not updating them separately runs you into the risk of not being able to evacuate successfully. Moreover, that evacuation process goes through the BBS API, which is another reason why the BBS needs to be updated separately from the cells as well.

Does that help?

anEXPer:
So: if BBS is immediately before the Brain, and the Brain is serial, won't
that gurantee the necessary sequence?

luan:
@anEXPer yeah that actually makes sense. I'm going to update the PR. Thanks!

luan:
There it is, I also cleaned up the history on the branch to make this look nicer on master. Thanks for the diligent review @anEXPer!

PR #106, opened 2017-03-29
ignore tmp

drnic:


dsabeti:
Hey @drnic. Happy to merge this, but I want to understand what it's for. Is there a specific reason to ignore `tmp`, or do you just see it as a good practice for people to make ignored local changes?

drnic:
Somewhere to store personal vars.yml and creds.yml and personal op patch files during experimentation.

Is there another way/place that cf-deployment & bosh-deployment suggest putting personal files?

dsabeti:
We've been using private git repos most recently -- you can see in [cf-deployment-concourse-tasks](https://github.com/cloudfoundry/cf-deployment-concourse-tasks) that it's our _currently_ supported approach there. Long-term, we'd prefer something more secure. Because CredHub is coming around the bend, we're waiting until we can use that to store creds for cf-deployment.

For _bosh-deployment_ creds/vars, you could try something like LastPass if you don't like using private repos.

Anyway, I think this makes sense so I'm going to merge it.

dsabeti:
Also @drnic, for future reference, PRs should be made against the `develop` branch.

PR #109, opened 2017-03-30
consistent 'default' name for all things: 'private' network now 'default'

drnic:
I've sanity tested this change by upgrading my cloud-config + deploying: all ok.

### DIFF
--- a/bosh-lite/cloud-config.yml
+++ b/bosh-lite/cloud-config.yml
@@ -4,7 +4,7 @@ azs:
 - name: z3
 compilation:
   az: z1
-  network: private
+  network: default
   reuse_compilation_vms: true
   vm_type: m3.medium
   workers: 6
@@ -18,13 +18,13 @@ disk_types:
 - disk_size: 100240
   name: 100GB
 networks:
-- name: private
+- name: default
   subnets:
   - azs: [z1, z2, z3]
     cloud_properties:
       name: random
     gateway: 10.244.0.1
-    range: 10.244.0.0/22
+    range: 10.244.0.0/19
     reserved:
     - 10.244.0.1
     static:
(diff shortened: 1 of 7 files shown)


drnic:
@dsabeti ETA for merging? 

https://github.com/cloudfoundry-community/docker-boshrelease/pull/79

jaresty:
Hey @drnic,

Thanks for this PR!  I'm looking forward to merging it.  We noticed three kinds of changes in this PR:

1. The rename of the private network to default
2. The addition of a public network to the bosh-lite cloud config
3. The addition of a default VM type to bosh-lite

We are fine with (1) but are unsure about (2) and (3).   Can you please explain the rationale for including those two additional kinds of changes in this PR?  Would you be open to making separate PRs for those two changes?

Thank you.

dsabeti:
@jaresty `bbl` actually gives you a [default vm type](https://github.com/cloudfoundry/bosh-bootloader/blob/master/cloudconfig/cloud_config_templates.go#L35) in your cloud-config. It makes sense to have the same thing in the bosh-lite cloud-config.

As for the update to add a public network, it's a deviation from what `bbl` gives you. @drnic, I think it's probably best to start by discussing the use case for a separate "public" network and adding it to `bbl`. Can we pull out that part of the PR for now?

jaresty:
@dsabeti That makes sense to me.  So the only outstanding question is:

@drnic: Are you OK with pulling the public network change out into its own PR?

drnic:
Yeah the public bit can come out.

drnic:
I've rebased against `develop` and updated the PR to merge into `develop`.

staylor14:
Hi @drnic 

I merged this into develop with a modification to the bosh-lite default network range.  It was originally a /22 network.  You increased its size to /19.  I moved it down slightly from there to /20 since this is what bbl provides (hooray eventual consistency!)

Thanks much!
--Steve

drnic:
@staylor14 lol I probably was having a brain fart moment and thinking I was reducing it. Thanks for fix.

PR #110, opened 2017-03-31
Add more certs for cc-uploader <-> diego mTLS

thausler786:
This adds the server certs required to mTLS communication between Diego's Reps and CC-Uploader. The Rep is already configured with certs of the same CA (the client cert is `diego_rep_client`).

The added properties will be required in an upcoming CAPI release.

https://www.pivotaltracker.com/story/show/142166993

Side note:

- We were unsure what conventions have been set for `common_name`

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -1013,6 +1013,9 @@ instance_groups:
             ca_cert: "((cc_bridge_cc_uploader.ca))"
             client_cert: "((cc_bridge_cc_uploader.certificate))"
             client_key: "((cc_bridge_cc_uploader.private_key))"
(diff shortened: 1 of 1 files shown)


anEXPer:
These properties seem weirdly scoped? It might be a little bit late in the process (and also slightly not my business) to be critiquing the spec structure of upstream releases, but `cc` being a peer to `ca_cert` inside `cc_uploader` seems confusing to me. I thought it was a typo at first.

Should these values be one layer down, inside something like `uploader_cert`?

aashah:
Cool, we updated the common name to be the hostname for cc-uploader.

PR #111, opened 2017-04-03
 Add TLS credentials for Diego Rep asset management.

bdshroyer:
* Asset uploads and downloads are now protected by TLS.
 * A single cf-deployment variable called rep-agent supplies credentials
   to the rep server and its internal asset uploaders/downloaders.

[finshes #141164981]


bdshroyer:
This branch didn't have a proper timestamp in the name, so we're replacing it with [PR #112](https://github.com/cloudfoundry/cf-deployment/pull/112).

PR #112, opened 2017-04-03
 Add TLS credentials for Diego Rep asset management.

bdshroyer:
Add TLS credentials for Diego rep's asset download/upload.

* Creates a new credential variable `diego-rep-agent` that gets used by the rep server and its internal asset uploader/downloader.

[finishes #141164981]

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -860,10 +860,14 @@ instance_groups:
           preloaded_rootfses:
           - cflinuxfs2:/var/vcap/packages/cflinuxfs2/rootfs
           require_tls: true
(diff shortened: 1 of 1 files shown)


bdshroyer:
Replaces [PR #111](https://github.com/cloudfoundry/cf-deployment/pull/111), which didn't have a proper timestamp in the branch name.

ChunyiLyu:
Hi @bdshroyer, thanks for contributing to cf-deployment! There were changes in the opsfiles that are also need to be made due to the name change. We have already pushed fixes to the develop branch, but just want to let you know.

@mingxiao && Chunyi

PR #122, opened 2017-04-11
cc-worker has properties to use routing api

thausler786:
When I am deploying with tcp routing, I want the cc worker to be able to speak to the routing api.  This commit adds the necessary properties to the cc-worker job.

Story:
https://www.pivotaltracker.com/story/show/143322377


thausler786:
@dsabeti this PR is necessary for the next CAPI release, when the fix for the RATS failure in 1.25.0 will come through.

PR #125, opened 2017-04-12
add experimental ops files for deploying locket component

crhino:
- Defaults database to mysql
- Uses diego_database_password for DB permissions
- Includes locket-postgres.yml file to work with postgres job
- Moves diego-bbs instance up in deployment order

[#143336051]

Signed-off-by: Stephen Carter <scarter@pivotal.io>


crhino:
Woops, forgot to move diego-bbs job up in the deploy order. Let me push that as well.

PR #127, opened 2017-04-13
Add vm agnostic names

genevievelesperance:
From the bbl team: https://www.pivotaltracker.com/story/show/140082675

cc @evanfarrar

### DIFF
--- a/bosh-lite/cloud-config.yml
+++ b/bosh-lite/cloud-config.yml
@@ -6,7 +6,7 @@ compilation:
   az: z1
   network: default
   reuse_compilation_vms: true
-  vm_type: m3.medium
+  vm_type: default
   workers: 6
 disk_types:
 - disk_size: 1024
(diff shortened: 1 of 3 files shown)


evanfarrar:
This is a change in direction and I wanted our team to open this as a discussion. We've been providing these names for a while in the BBL generated cloudconfig for AWS and GCP and hope to avoid making the operator/core developer choose among names from another IaaS they might not be familiar with. This is part of the reason that BOSH provides re-nomination or VM type names instead of directly addressing them: so that VM types can use names that make sense to the operator.

If an operator wants to scale a job vertically these names give an intuitive indication of size. One day in the glorious future BOSH may allow a concrete definition of size  to be backronymed into IaaS naming schemes but this convention will serve as a bridge (e.g. 8 cpus and 16gigs of memory in the manifest becomes m4.xlarge without additional cloud-config declarations).

This may not cover every single reference to a VM name using the AWS scheme but it should be enough to get the ball rolling towards that end.

dsabeti:
Howdy @genevievelesperance && @evanfarrar.

I think we've settled on some ideas around vm types for cf-deployment, and I've written [a story to use iaas-agnostic VM types](https://www.pivotaltracker.com/story/show/149037813) based the new types in `bbl`'s cloud-config, which vm types we'd prefer to use. One important decision we made was not to use the `default` vm type.

Do we think it makes sense to close this PR?

genevievelesperance:
Yep! Thanks!

PR #128, opened 2017-04-13
Don't hard code stemcell version, please

drnic:
Since bosh cannot download the stemcell, it makes documentation awkward (and eternally wrong/needing updates) if we force a specific version of the stemcell here. Can we please not hardcode the stemcell version?

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -1550,4 +1550,4 @@ releases:
 stemcells:
 - alias: default
   os: ubuntu-trusty
-  version: "3363.14"
+  version: latest


drnic:
See comments of http://www.starkandwayne.com/blog/running-cloud-foundry-locally-on-bosh-lite-with-bosh2/ for the issues coming from the hardcoded version number

dsabeti:
Hey @drnic. Thanks for this. A few things:
- The stemcell version is [automatically updated](https://release-integration.ci.cf-app.com/teams/main/pipelines/cf-deployment/jobs/update-releases/builds/21) every time a new stemcell is released. Take a look at the `develop` branch, and you'll see that it has the latest version (3363.15). We just need to cut a new version of cf-deployment -- I'll get to it today. (Cutting new versions updates the `master` branch). This is valuable feedback about the release schedule for cf-deployment, especially when it comes to stemcells.
- The point of encoding the stemcell version (rather than using `latest`) is to make sure that a deployer knows which stemcell they're actually going to deploy. Especially in automated workflows, it's very easy to think that you're getting the right stemcell when the `latest` keyword is actually obscuring the version of the stemcell that gets deployed.

Let me know if you have any other thoughts or questions about this.

rkoster:
Has the ['And in the future'](https://github.com/cloudfoundry/bosh-notes/blob/master/finished/asset-fetching.md#stemcells) stuff already been implemented? Because having something like that would also solve this problem.

rkoster:
We are using https://github.com/cloudfoundry/bosh-deployment-resource, and currently do:
```yaml
resources:
  - name: stemcell
    type: bosh-io-stemcell
    source:
      name: bosh-vsphere-esxi-ubuntu-trusty-go_agent
      version_family: "3363.14"
  - name: cf-deployment
    type: git
    source:
      branch: master
      uri: https://github.com/cloudfoundry/cf-deployment.git
  - name: bosh-deploy
    type: bosh2-deployment
    source:
      deployment: cf
      target: {{bosh_target}}
      client: {{bosh_client}}
      client_secret: {{bosh_client_secret}}
      ca_cert: {{bosh_ca_cert}}

jobs:
  - name: deploy-cf
    plan:
      - aggregate: 
          - get: cf-deployment
          - get: stemcell
      - put: bosh-deploy
        params:
          manifest: cf-deployment/cf-deployment.yml
          stemcells:
          - stemcell/*.tgz
          ops_files:
          - cf-deployment/operations/use-compiled-releases.yml
          - cf-deployment/operations/scale-to-one-az.yml
          vars:
            system_domain: {{system_domain}}
```

The hardcoding of the stemcell version is something we would really like to get rid of.

dsabeti:
@rkoster Thanks for chiming in.

Something I forgot to mention in my first response is that we also use cf-deployment.yml to include the "blessed" combination of releases and stemcells -- if you're using master, then you're getting a combination that's passed all of our tests. It sort of replaces and expands the [compatibility matrix](https://github.com/cloudfoundry/diego-cf-compatibility) that we've used for tracking integration of CF and Diego. Using `latest` dilutes that guarantee.

Could you explain a little more how the features described in the document about fetching assets would solve this problem? From what I can tell, that example still has versions hard-coded in the manifest. (And I'm not sure if the features described there have been implemented. @cppforlife?)

As for automation, I think you bring up an interesting use case. If you want to use a concourse resource for your stemcell, you need a way to encode in the manifest the stemcell version as fetched by the resource. `latest` would achieve that, but I think there are other solutions as well.
- The simplest solution that I can think of is supporting an ops file for using the latest stemcell.
- If your primary concern is Concourse, you might also consider using [this task](https://github.com/cloudfoundry/cf-deployment-concourse-tasks/tree/master/bosh-upload-stemcell-from-cf-deployment), which uploads whichever stemcell is encoded in the manifest. This lets you remove the stemcell resource from your pipeline.

rkoster:
The problem as described by @drnic is one of writing a blog post which get out of date (because of `bosh upload-stemcell ...`), but this also applies to concourse pipelines (could be solved with the above task). 

Hardcoding blessed version is really neat. But for some usecases, say a blog post. You just want bosh deploy to just work. Maybe having an `latest-stemcell.yml` ops file similar to what they have in the [cf-mysql-deployment](https://github.com/cloudfoundry/cf-mysql-deployment/blob/master/operations/latest-versions.yml#L8-L13) repo would help with that.

But the ideal solution would be a hash with blessed stemcells for all cpis (as from described in bosh-notes). Because `bosh deploy` would just work 😄 (whiteout having to think about stemcells).

dsabeti:
I see. You're saying that if the manifest itself included the urls for the stemcells on each CPI, and the BOSH cli automatically uploaded the right one, then deployers would never have to consider stemcell versions at all, because the whole workflow would be seemless.

Let me get in touch with the BOSH team about that feature.

rkoster:
Yes indeed. Which would also make the above concourse pipeline infrastructure agnostic.

dsabeti:
Hey @rkoster. I spoke with @cppforlife and it looks like BOSH isn't planning on implementing the feature to specify stemcell urls in the manifest.

In the meantime, I've written a story to provide the ops file you're asking for: https://www.pivotaltracker.com/story/show/143906715. Between that story and the Concourse task that I linked to, do you have what you need?

drnic:
David, sorry for not following up earlier.

Can we switch to 3363.latest version instead?

________________________________

dsabeti:
Hey @drnic. I'm not sure that I like the idea of using the `X.latest` model. I think we'd miss out on the benefits of _both_ models (using `latest` vs. hard-coding the version): If there's a major version bump, the deployer will still have the manual step of uploading the stemcell, and the manifest becomes less deterministic.

Do you think that an ops-file for using the latest stemcell, as suggested by @rkoster, would suffice? Is the issue that you don't want a first-time deployer to have to add in an ops-files to their deploy?

rkoster:
@dsabeti yes the story as written seems sufficient for the first time deployer usecase.

dsabeti:
Hey @drnic. I think it's best to close this PR. Like I explained, I prefer to let the manifest continue to encode compatible versions of stemcells and releases, but I'm happy to prioritize work to simply allow users to go with the latest stemcell.

PR #129, opened 2017-04-13
Expose consul as cross deployment link

drich10:
As an operator I would like to be able to run the `consul_agent` in other deployments without needing to configure IP information. Exposing the cf consul_server link allows me to use links to solve this issue.


drich10:
Duplicate of #119 closing

PR #131, opened 2017-04-19
add a 'default' disk_type

drnic:
cf-deployment's bosh-lite/cloud-config.yml will be a defacto cloud-config; to allow all 3rd party manifests to "just work", we'll includes a 'default' for all cloud-config entities

From my "BOSH 2.0" talk in Toronto and at today's CAB call:

screen shot 2017-04-19 at 9 26 35 am

Sorry I missed this in https://github.com/cloudfoundry/cf-deployment/pull/109

### DIFF
--- a/bosh-lite/cloud-config.yml
+++ b/bosh-lite/cloud-config.yml
@@ -9,6 +9,8 @@ compilation:
   vm_type: m3.medium
   workers: 6
 disk_types:
+- disk_size: 10240
+  name: default
 - disk_size: 1024
   name: 1GB
 - disk_size: 5120


dsabeti:
Hey @drnic. This seems like a reasonable change, but I noticed that nothing in cf-deployment would actually use the `default` disk type in this change. Would it make sense to model the `default` disk type after whichever disk type is most commonly used in cf-deployment and then update the manifest to use `default`?

Also, `bbl` doesn't currently create a disk type called `default`. @evanfarrar (cc @christianang), should `bbl` create a disk type called `default` in its cloud-config?

drnic:
As discussed in CAB, there can only be one cloud-config per BOSH, and cf-deployment will be the canonical source for many ppl. So default might not be used by cf-deployment but it will satisfy the pattern of expectation by deployments that collocates on the same bosh.

________________________________

evanfarrar:
I think one concern would be that cf-deployment might break this expectation that a default disk exists at some date in the future, accidentally, if nothing uses it. If we could change a fair number of the VMs in cf-deployment on multiple IaaS refactored to use default, then that expectation would be safer.

This, however, would break in CI until BBL also started providing a disk type named `default` on the BOSH directors that @dsabeti 's team runs CATS against.

@dsabeti I'm happy to see that change through with a later PR to utilize the disk type in `cf-deployment.yml` once BBL is providing it. I think it would be fine to accept this PR in the mean time to start socializing this idea.

@drnic for bosh-lite it is fine to have a name-only disk-type, but a real IaaS will want at least a number of gigabytes to allocate. For your own BOSH releases that you personally maintain, what size would you expect `default` to be? I see 5gb used the most (twice) in `cf-deployment.yml`, is that enough?

drnic:
My primary goal is to get a consistent base pattern across all cloud-config that ppl share. cf-deployment is the heart of the cf community so it is the basis from which de facto patterns emerge.

Can you have a quick chat with @cppforlife and perhaps agree on what minimal set of names/labels we should all put in each cloud-config?

I'm coming into SF office tomorrow/Thurs afternoon. Happy to help settle on something then if you'd like an extra set of opinions.

My only goal is to settle somewhere, then I can help PR every other cloud-config I find on the internet until they are all minimally interchangeable.

________________________________

dsabeti:
Cool. It sounds like
- I can prioritize this PR. 
- @evanfarrar and @drnic should discuss what makes sense for a default value.
- @evanfarrar can write a story to add a default disk type to `bbl`
- cf-deployment can start using the `default disk type once `bbl` provides it.

Did I get that right?

PR #132, opened 2017-04-19
Use minio as cloud_controller blobstore

rkoster:
This PR replace the webdav blobstore with a 4 node minio cluster
minio_browser

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -175,6 +175,54 @@ instance_groups:
   - name: metron_agent
     release: loggregator
     properties: *metron_agent_properties
+- name: router
+  azs:
+  - z1
+  - z2
+  instances: 2
+  vm_type: m3.medium
+  vm_extensions:
+  - cf-router-network-properties
+  stemcell: default
+  update:
+    max_in_flight: 1
+    serial: true
+  networks:
+  - name: default
+  jobs:
+  - name: consul_agent
+    release: consul
+    consumes:
+      consul: {from: consul_server}
+    properties:
+      consul:
+        agent:
+          services:
+            gorouter: {}
+  - name: gorouter
+    release: routing
+    properties:
+      router:
+        ssl_skip_validation: true
+        enable_ssl: true
+        ssl_cert: "((router_ssl.certificate))"
+        ssl_key: "((router_ssl.private_key))"
+        status:
+          password: "((router_status_password))"
+          user: router-status
+        route_services_secret: "((router_route_services_secret))"
+        tracing:
+          enable_zipkin: true
+      uaa:
+        clients:
+          gorouter:
+            secret: "((uaa_clients_gorouter_secret))"
+        ca_cert: "((uaa_ssl.ca))"
+        ssl:
+          port: 8443
+  - name: metron_agent
+    release: loggregator
+    properties: *metron_agent_properties    
 - name: mysql
   azs:
   - z1
@@ -716,54 +764,6 @@ instance_groups:
         url: https://uaa.((system_domain))
         ssl:
           port: 8443
-- name: router
-  azs:
-  - z1
-  - z2
-  instances: 2
-  vm_type: m3.medium
-  vm_extensions:
-  - cf-router-network-properties
-  stemcell: default
-  update:
-    max_in_flight: 1
-    serial: true
-  networks:
-  - name: default
-  jobs:
-  - name: consul_agent
-    release: consul
-    consumes:
-      consul: {from: consul_server}
-    properties:
-      consul:
-        agent:
-          services:
-            gorouter: {}
-  - name: gorouter
-    release: routing
-    properties:
-      router:
-        ssl_skip_validation: true
-        enable_ssl: true
-        ssl_cert: "((router_ssl.certificate))"
-        ssl_key: "((router_ssl.private_key))"
-        status:
-          password: "((router_status_password))"
-          user: router-status
-        route_services_secret: "((router_route_services_secret))"
-        tracing:
-          enable_zipkin: true
-      uaa:
-        clients:
-          gorouter:
-            secret: "((uaa_clients_gorouter_secret))"
-        ca_cert: "((uaa_ssl.ca))"
-        ssl:
-          port: 8443
-  - name: metron_agent
-    release: loggregator
-    properties: *metron_agent_properties
 - name: diego-brain
   azs:
   - z1
--- a/operations/use-minio.yml
+++ b/operations/use-minio.yml
@@ -16,9 +16,20 @@
     persistent_disk_type: 100GB
     networks: [ name: default ]     
     jobs:
+      - name: consul_agent
+        release: consul
+        consumes:
+          consul: {from: consul_server}
+        properties:
+          consul:
+            agent:
+              services:
+                blobstore:
+                  check: {}
       - name: minio-server
         release: minio
         properties:
+          port: 80
           credential:
             accesskey: "((minio_accesskey))"
             secretkey: "((minio_secretkey))"
@@ -28,7 +39,7 @@
           route_registrar:
             routes:
               - name: blobstore
-                port: 9000
+                port: 80
                 registration_interval: 20s
                 tags:
                   component: blobstore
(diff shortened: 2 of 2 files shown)


rkoster:
Hmmm... not yet ready.. just found
```
2017-04-19T21:49:37.09+0200 [STG/0] OUT Downloading ruby_buildpack failed
```
probably due to cert nog being trusted.

dsabeti:
Haha, I was about to follow up and ask how you're testing this.

Overall, I think this is an interesting change to include in cf-deployment. I do want to make sure we get test coverage of it somewhere, and I think there's some more general discussion about supporting community-contributed ops-files.

Let me know when you're figured out the issue you mentioned in your comment, and I'll keep you updated on how we'll try to support this.

rkoster:
@dsabeti with the above commit I was able to do a clean cf deploy. The smoke tests work a well.
For this to work I switched to using consul and configure fog to go over `http`.
The downside of using an internal route is that the following won't work:
```
cf curl "/v2/apps/$(cf app cf-env --guid)/droplet/download"
FAILED
Error creating request:
Error performing request: Get http://blobstore.service.cf.internal/cc-droplets/6e/a4/6ea4e04c-491d-4411-864f-65d7cba6a422/3b4ce43ccb676b9baab54d66d6a0a3d49d9b48c1?AWSAccessKeyId=z9k42Kpf54MfMj6AUrvD&Signature=Dk8mHFXXF%!B(MISSING)DZsuAnIo0vLIdDWoc%!D(MISSING)&Expires=1492609341: dial tcp: blobstore.service.cf.internal: no such host
TIP: If you are behind a firewall and require an HTTP proxy, verify the https_proxy environment variable is correctly set. Else, check your network connection.
```

But this should be fixed once `cf-deployment` starts using [the bits-service-release](https://github.com/cloudfoundry-incubator/bits-service-release)

dsabeti:
Hey @rkoster. I'm a bit confused about this last bit where `cf curl /v2/apps/:guid/droplet/download` doesn't work. The webdav blobstore [also uses consul and an internal route](https://github.com/cloudfoundry/cf-deployment/blob/master/cf-deployment.yml#L454-L462), but users can still download droplets. I think the more important thing is whether or not the blobstore registers a public route; the webdav blobstore [registers a route with the router](https://github.com/cloudfoundry/cf-deployment/blob/master/cf-deployment.yml#L476-L487). Would the same work for minio?

rkoster:
@dsabeti having a `private_endpoint` and a `public_endpoint` is a `webdav` only thing [as far a I know](https://github.com/cloudfoundry/cloud_controller_ng/blob/master/bosh/jobs/cloud_controller_clock/spec#L227-L232). I also configure a public route the router but there is no way to configure `cloud_controller_ng` to use it. 

It looks like the bit-service will introduce [private and public urls](https://github.com/cloudfoundry-incubator/bits-service-release/blob/master/jobs/bits-service/spec#L32-L36) for [fog based blobstores](https://github.com/cloudfoundry-incubator/bits-service-release/blob/master/jobs/bits-service/spec#L53-L55).

It would be posible to get the same public/private effect with a [dns-alias](https://github.com/cppforlife/dns-release/blob/master/manifests/runtime.yml#L14) but this would require people to use the `dns-release`.

But since the smoke tests work with current setup, and with the prospect of the bit-service fixing the public/private I don't think the added `dns-release` complexity would be worth it.

dsabeti:
I see. It looks like the fog "support" for public/private urls [is to use the same url for both](https://github.com/cloudfoundry/cloud_controller_ng/blob/master/lib/cloud_controller/blobstore/fog/fog_blob.rb#L15-L21). Does it make sense instead to register a public route only (i.e., not use consul at all) and have the Cloud Controller use that instead?

rkoster:
@dsabeti that is indeed how I started, but then there is a ordering issue on a clean deploy.
Because the cloud_controller needs a blobstore because it uploads the buildpacks in it's post-start, but at that point the routers are not yet up and running.

The second problem with this aproach is [self signed blobstore certs and buildpack downloads](https://github.com/cloudfoundry/cf-deployment/pull/132#issuecomment-295417296). This should however work fine if you are using valid certs on you load balancer. But this introduces an extra external requirement.

dsabeti:
Ok, with regards to the first problem, you may need to play around with the deploy order.

For the second problem, take a look at this property: https://github.com/cloudfoundry/diego-release/blob/develop/jobs/rep/spec#L115-L117

Overall, I'm happy to take this PR, assuming two things:
1. These problems all get ironed out.
2. I'll probably place it in a directory called something like `operations/community` and I'll list you (or whoever else you want) as the primary supporters of the ops file. I don't want to get into the business of testing every possible permutation of configuration options, so for things like this I'd like to defer to whoever submitted the ops-file in the first place.
How does this sound?

rkoster:
@dsabeti I was able to deploy succesfull with the above changes. Also the CATs test suite passed.
I'm happy to be listed as primary supporter of this opps file, or maybe better to list Stark & Wayne.

dsabeti:
Great! I've prioritized the story so that the engineering team can take a look at the change.

dsabeti:
Hey @rkoster, I wanted to give you an update on the status of this PR. 

Someone pointed out that a complicating factor in this is the potential use of the Routing API to register routes, rather than NATS. Such a configuration would require yet another rearrangement of the instance groups in cf-deployment. I'm doing some research now to discover the deploy-time dependencies of the various jobs, to see if we can make better sense of how this ops-file can re-order the instance groups. It's possible that we'll change colocation strategy as a result.

Anyway, there are two options for this PR:
1. We continue to leave it open while we do the work to come up with a more meaningful deploy order. Once that's ready, we can merge the PR.
2. Merge the PR now so that it's available for use, while we do the research in the background. When that research is done, it's very likely that we'll make changes that are incompatible with the ops-file, so you'd have to come in and update based on the changes that we've made.

How do you want to proceed?

rkoster:
Thanks for the update, I rather keep it open. So people get the whole context. If they want to use it, they can copy the file form this PR to their own ops-files repo is use it from there.

abhilash07:
@rkoster i have tried minio for cf blobstore. But unfortunately i encountered below error.

image

[cf-deployment.txt](https://github.com/cloudfoundry/cf-deployment/files/1990028/cf-deployment.txt)

[deployment-vars.txt](https://github.com/cloudfoundry/cf-deployment/files/1990030/deployment-vars.txt)

after checking the cloud_controller_ng post-sart log i found this error.

"rake aborted!
[2018-05-04 05:04:41+0000] CloudController::Blobstore::BlobstoreError: Connection refused - connect(2) for 10.244.0.34:443 (Errno::ECONNREFUSED)"
[post-start.stderr.log](https://github.com/cloudfoundry/cf-deployment/files/1990029/post-start.stderr.log)

Do you any idea about this error.

Thanks in advance.

rkoster:
@abhilash07 check if your load balancer is up, and your gorouters + nats are all running. Also make sure minio is up and running. If some part is not deployed yet you will have to change the deployment order in the manifest.

abhilash07:
@rkoster thanks for you responce you mean before cc deploying we have to deploy haproxy-loadbalancer and go router along with nats and minio ???

as you see my manifest my go router and haproxy can be active after cc-api is active. shoukd i want to change the order ???

For Example:
                 1) nats
                 2) minio
                 3) go router
                 4) haproxy-loadbalancer
                 5) api (cloud-controller) ???

Thank in advance

rkoster:
yes, alternatively you can switch to using bosh dns to reference the internal url of the minio, but this way you won't be able to download droplets via the CC api.

abhilash07:
@rkoster thanks for your information, you said bosh dns. If i am not wrong the bosh dns should to configured at bosh-director (Micro-bosh) level right ???

abhilash07:
@rkoster thanks for your suggestions, with out dns it is sucessfull. Soon i will pull and push minio as blobstore to cf-deployment branch.

gberche-orange:
Having this PR merged would be quite useful to us at orange, and will protect against future inconstencies that this unmerged PR might have with next changes to [use-s3-blobstore.yml](https://github.com/cloudfoundry/cf-deployment/blob/master/operations/use-s3-blobstore.yml) (e.g. blobstore config entry list gets modified). We experienced a related production incident documented at https://github.com/orange-cloudfoundry/paas-templates/issues/91

@heyjcollins  @dsabeti I understand from https://github.com/cloudfoundry/cf-deployment/pull/132#issuecomment-305665271 that this PR was waiting for the relint team to complete the rearrangement of the instance groups in cf-deployment, so that minio server can be registered using routing API. Is it still the case ? Is there something the community can do to help here ?

Thanks in advance,

Guillaume.

heyjcollins:
Hey @gberche-orange - I'll review this with the RelInt team this week and let you know if we're prepared to roll forward with it.

davewalter:
Closing due to inactivity. Please feel free to reopen if desired.

PR #133, opened 2017-04-19
Cell presence in locket

jvshahid:
This pr adds a new ops file to configure the cells to use locket for their cell presence instead of consul

[#142583213]

Signed-off-by: Swetha Repakula <srepaku@us.ibm.com>


staylor14:
Hi @jvshahid,

Can this not just ride along in the existing locket ops file, or do you need to be able to control this independently from the settings there?

Thanks!
--Steve && @ChunyiLyu

jvshahid:
talked to @ematpl and he agreed to merging the two files. I force pushed the change btw.

PR #138, opened 2017-04-21
add experimental ops files for local route emitter emitting TCP routes

crhino:
There are a number of dependencies for these ops files.

- Both depend on operations/tcp-routing-gcp.yml
- operations/enable-local-route-emitters.yml for
enable-local-route-emitters-tcp.yml
- windows-cell.yml and local-route-emitter-windows ops files for windows
tcp ops file.

[This PR](https://github.com/cloudfoundry/cf-deployment/pull/130) should take care of the local-route-emitter dependencies.

This also does not currently work on bosh-lite out of the box, since the
bosh-lite cloud config does not have the VM extension
`internet-required`

[#142583351]

Signed-off-by: Chris Piraino <cpiraino@pivotal.io>

### DIFF
--- a/operations/experimental/enable-local-route-emitter-tcp-windows.yml
+++ b/operations/experimental/enable-local-route-emitter-tcp-windows.yml
@@ -1,0 +1,13 @@
+---
+- type: replace
+  path: /instance_groups/name=windows-cell/jobs/name=route_emitter_windows/properties/tcp?/enabled?
+  value: true
+
+- type: replace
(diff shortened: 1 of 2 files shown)


crhino:
Currently discussing whether to just merge these two ops files into tcp-routing-gcp.yml instead. Consider this a work-in-progress.

crhino:
Decision is to proceed with adding these experimental files on top of `tcp-routing-gcp.yml` and merge them when routing team no longer wishes to test the global tcp emitter.

staylor14:
Merged #144130355

PR #139, opened 2017-04-21
Be more precise on where to find bosh-lite CA cert

jenspinney:
Before this change, the README referenced `$PWD` without mentioning which directory the user should be in when running that comand.


dsabeti:
Thanks @jenspinney!

PR #143, opened 2017-04-25
Add haproxy support

rkoster:
This PR adds https://github.com/cloudfoundry-incubator/haproxy-boshrelease as a load balancer for environments without IaaS provided load blancers.

### DIFF
--- a/operations/use-haproxy.yml
+++ b/operations/use-haproxy.yml
@@ -1,0 +1,43 @@
+# Install HA Proxy
+- type: replace
+  path: /releases/-
+  value:
+    name: haproxy
+    version: 8.0.12
+    url: git+https://github.com/cloudfoundry-incubator/haproxy-boshrelease
+    
+- type: replace
+  path: /instance_groups/-
+  value:
+    name: haproxy
+    azs: [z1, z2]
+    instances: 2
+    stemcell: default
+    vm_type: m3.medium
+    networks: [ name: default ]
+    jobs:
+      - name: consul_agent
+        release: consul
+        consumes:
+          consul: {from: consul_server}
+      - name: keepalived
+        release: haproxy
+        properties:
+          keepalived:
+            vip: ((haproxy_static_vip))
+      - name: haproxy
+        release: haproxy
+        properties:
+          ha_proxy:
+            backend_servers: [ gorouter.service.cf.internal ]
+            ssl_pem: "((router_ssl.certificate))((router_ssl.private_key))"
+            tcp:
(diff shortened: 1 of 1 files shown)


dsabeti:
Hey @rkoster. Sorry for the delayed response to this.

I'm pretty excited to see the new haproxy release in action, so I'm glad you've submitted this. One thing we were thinking about was if it made more sense to include this as a separate deployment manifest rather than an ops-file. The thinking is that it more closely resembles the IaaS load balancers that are deployed and operated separately. Also, you get the added benefit of updating the load balancer and CF instance separately, so operators have more control of availability. 

What do you think? Could this be a separate deployment manifest? If so, we'd gladly host the manifest in this repo.

rkoster:
@dsabeti with the [new haproxy release v8.1.1](https://github.com/cloudfoundry-incubator/haproxy-boshrelease/releases/tag/v8.1.1) there is links support. I have updated the PR to use this. Also I have replace the smoke-tests with haproxy (and readded the smoke-tests as the last instance_group) as a workaround for https://github.com/cppforlife/go-patch/issues/6. 

This is needed when [wanting to use minio](https://github.com/cloudfoundry/cf-deployment/pull/132). An alternative would be adding a placeholder instance_group for `load_balancer` to [`cf-deployment.yml`](https://github.com/cloudfoundry/cf-deployment/blob/master/cf-deployment.yml).

With regards to putting haproxy in it's own deployment.. I would rather keep it in the cf deployment. Because for environments without external loadbalancers CATS would not succeed without it. Also now that the haproxy-release is using linking, we would have to make the consumed links shared.

dsabeti:
Hey @rkoster. 

I was going to argue in favor using cross-deployment (i.e. shared) links, but it was recently pointed out to me that shared links only get updated when you re-deploy the _consumer_ of the link. In this case, if the haproxy is in a separate deployment and consumes the `http-router` link, you'll have to re-deploy the haproxy every time you re-deploy the routers -- including scaling them up or down. This creates a pretty tight coupling that would be obscured by the fact that haproxy is in a different deployment. I'm leaving this explanation here in the PR for future reference, in case anyone wonders whether we considered using separate deployments.

I'll prioritize the story now.

selzoc:
Merged this, but on our develop branch, with https://github.com/cloudfoundry/cf-deployment/commit/91b6b2f617cc476b80842e17c48422c816909b59

See the ops file: https://github.com/cloudfoundry/cf-deployment/tree/develop/operations/community

PR #147, opened 2017-05-01
Add ops file to bypass CC bridge

tcdowney:
add ops file to bypass CC bridge

- this enables the temporary flags that enable mTLS communication between
    CC and Diego
- as a result, the stager, nsync, and tps-listener functionalities are now handled by
    CC, so we can remove these jobs from the cc-bridge template
- these flags are turned on in the api, cc-worker, and cc-clock vms

[#139962927]

### DIFF
--- a/operations/bypass-cc-bridge.yml
+++ b/operations/bypass-cc-bridge.yml
@@ -1,0 +1,85 @@
+---
+# cloud_controller_ng
+- type: replace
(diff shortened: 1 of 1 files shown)


tcdowney:
We realized that putting the YAML anchors on the `cc.diego` key makes this ops file less composable than it could be.  We're going to tweak it a bit so that it doesn't unintentionally wipe out any other config that could be under that key.

PR #150, opened 2017-05-05
Rename the alias for the old consul link

davewalter:
Hi,

Sorry to do this to y'all, but apparently our local testing of our last PR wasn't thorough enough to catch a problem with a collision between the alias we had used for the old consul link and the name of one of the new consul links. This PR changes the alias used for the old single link from consul-release. We have carefully tested a deployment with this change and confirmed that it works as desired.

Thanks!
Dave

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -50,12 +50,12 @@ instance_groups:
   - name: consul_agent
     release: consul
     consumes:
-      consul: {from: consul_server}
+      consul: {from: consul_link}
       consul_common: nil
       consul_server: nil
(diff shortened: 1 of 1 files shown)


ChunyiLyu:
Hey @davewalter, we had to `ag` for other instances of the old link name to fix all our ops files to work with this change. Try to keep ops files in mind when making high surface-area changes like this. Thanks for the PR!

davewalter:
Thanks for taking care of this for us. We had a rough time with this change overall and this definitely slipped through the cracks. Sorry about that.

PR #154, opened 2017-05-16
Adds `consumes.nats` to `route_emitter` job

coreyti:
This addition makes it simple to provide an ops file that updates only
the `deployment` key, when deploying the `diego-cell` instance group as
a separate deployment, with cross-deployment links.

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -912,6 +912,8 @@ instance_groups:
           bbs: *diego_bbs_client_properties
   - name: route_emitter
     release: diego
+    consumes:
+      nats: {from: nats}
     properties:
       diego:
         route_emitter:


ChunyiLyu:
Hey @coreyti, thanks for the PR.

We've got a couple of thoughts about this.

Firstly, how does this simplify adding an ops-file to target this value? A value need not be present in the manifest in order for an ops-file to target it. The `?` character can be put in the `path` of the operation, which will cause a path to be added if it's not found in the source manifest, and it doesn't seem substantially more complicated to add the whole "consumes" section rather than setting just the deployment key. We have an intention, in cf-d, of avoiding "target" configuration, default stuff that is there just to be changed, and this seems like it falls into that category, unless we're misunderstanding.

HOWEVER.

Maybe we could include the "cross deployment" link by default? That is, we could consume the link explicitly from the "cf" deployment, from within that deployment? Assuming this works at all, this would mean that cell configuration extracted from cf-d and deployed under a different name would come, out of the box, with a default that consumes the link from the main cf-d deployment.

We've not tested that you can consume a cross-deployment link from within that deployment, though.

Thoughts?

selzoc:
@coreyti bumping this

coreyti:
@ChunyiLyu, @anEXPer, @selzoc,

Despite my intro comment, in my initial exploration into separating out the Diego cells, I built a small CLI that would extract a given instance group from a deployment manifest, into a separate manifest. Using that approach, it was simpler to find any instances of `consumes`, and add only the `deployment` pair.

That approach works well if everything else is already specified. It's true that, aside from providing a hint, this change doesn't provide much value when ops files can handle the change just fine.

That said, I like the suggestion of including the self-referential `platform` in the base manifest. I believe this does work fine.

anEXPer:
Alright, thanks for the context, that makes sense. We're going to close the PR for now (following our "no unnecessary config" heuristic), but we'll keep this stuff in mind as we zero in on an approach for supporting multi-deployment production configurations of cf-d.

PR #164, opened 2017-06-13
Disable TLS validation on cc clock so it can function with Routing API

tcdowney:
- the cc clock calls out to the routing api while doing the diego_sync
  job
- this is similar to https://www.pivotaltracker.com/n/projects/1382120/stories/146495795

[#147064255]

Signed-off-by: Tushar Singal <tsingal@pivotal.io>

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -975,6 +975,8 @@ instance_groups:
       system_domain: "((system_domain))"
       app_domains: *app_domains
       routing_api: *routing_api
+      ssl:
+        skip_cert_verify: true
       uaa:
         ca_cert: "((uaa_ssl.ca))"
         clients:


zrob:
not a fan of setting insecure mode as the default

it seems we already do this in the cc and cc_worker jobs....this also seems bad

i'd prefer we were in the secure mode by default with an opt-out ops file

tcdowney:
Will close this and put out a PR that does what @zrob suggested. :1st_place_medal:

dsabeti:
@zrob @tcdowney Sounds good, but for what it's worth, we do have this other (currently unprioritized) track of work to use bosh trusted certs to enable TLS validation everywhere: https://www.pivotaltracker.com/story/show/145086337

PR #176, opened 2017-07-10
Updated README

tinygrasshopper:
Removed non existent entry in the table of contents


dsabeti:
Good catch! Thanks for the PR.

PR #181, opened 2017-07-20
Add secure-service-credentials ops file

ishustava:
This PR adds an ops file, which enables secure service credentials delivery with CredHub.

### DIFF
--- a/operations/experimental/secure-service-credentials.yml
+++ b/operations/experimental/secure-service-credentials.yml
@@ -4,9 +4,9 @@
   path: /releases/-
   value:
     name: credhub
-    url: https://bosh.io/d/github.com/pivotal-cf/credhub-release?v=1.1.2
-    version: 1.1.2
-    sha1: 1ceacbbcb877a9b591a60a5f8f6f868d3147545a
+    url: https://bosh.io/d/github.com/pivotal-cf/credhub-release?v=1.2.0
+    version: 1.2.0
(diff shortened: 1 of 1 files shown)


pivotal-danjahner:
This ops file is missing a property to enable CC > Diego communication. This should be added - 

```
- type: replace
  path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/cc/diego?/temporary_local_apps
  value: true
- type: replace
  path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/cc/diego?/temporary_local_tasks
  value: true
- type: replace
  path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/cc/diego?/temporary_local_staging
  value: true
```

ishustava:
@pivotal-danjahner

What we haven't mentioned is that this ops file has a dependency on `bypass-cc-bridge.yml` ops file in cf-deployment which accomplishes that. Whiteout that ops file CATs were not passing. We have [documented it in the `credhub-release` repo](https://github.com/pivotal-cf/credhub-release/blob/master/docs/secure-service-credentials.md)

dsabeti:
@ishustava @pivotal-danjahner Just a heads up, when we merge this PR, we're going to add functionality to our pipeline to update the release version automatically when new credhub releases are shipped. Let us know if there's a problem with that approach.

pivotal-danjahner:
Sounds good.

PR #191, opened 2017-08-02
Support cf-deployment on Azure

bingosummer:
The purpose of the code changes is to support cf-deployment on Azure.
With azure.yml and the cloud-config.yml (described in azure-deployment-guide.md), users can deploy cf-deployment on Azure.
Since bosh-bootloader (bbl) doesn't support Azure for now, BOSH CLI v2 is used to setup DIRECTOR, and we create load balancers via Azure CLI.
After bbl supports Azure, bbl can generate the cloud-config.yml for Azure, and create Azure load balancers.

Additionally, use-azure-storage-blobstore.yml is to use Azure Storage as the blobstore of CF.


dsabeti:
Hi @bingosummer, thanks for this PR! We're really excited to have Azure support for cf-deployment. I've prioritized the story to merge this PR in our backlog.

Since this is a pretty important feature, and includes some docs (which is great), we're going to review the PR pretty carefully and potentially make some edits. We'll follow the instructions to verify it on our own Azure environment as well.

bingosummer:
Thanks a lot for reviewing. Could you please share the status of the reviews? If there are any issues about the PR, please let me know.

dsabeti:
Hey @bingosummer, we haven't started reviewing the PR. It's in our backlog of work -- you can see where it is here: https://www.pivotaltracker.com/story/show/149872014

We're going to try to get to it some time next week.

dsabeti:
Hi @bingosummer, we've finally merged this PR. I did want to let you know of a few changes we made.
1. We decided to consolidate deployment instructions into the README, so you'll notice that both the Azure and GCP deployment guides are no longer present. This seems like the right time to make our documentation less IaaS-specific. As Azure support makes its way into `bbl`, we feel pretty confident that everyone will have a similar UX.
2. We've also started using IaaS-agnostic vm_types in the cf-deployment manifest, so we no longer need the `azure.yml` ops-file.

Thanks again for this contribution!

bingosummer:
Thanks @dsabeti and other folks to review the PR.
I notice that `tcp-router` is by default inlined in cf-deployment.yml. Normally, would it need a different load balancer? If so, I should create 2 lbs before deployment. Correct?

dsabeti:
@bingosummer, that correct. And it would need to open up ports  1024-1123. Keep in mind that `bbl` will create such a load balancer soon.

PR #192, opened 2017-08-02
Remove duplicate python-buildpack

jsievers:


jsievers:
this PR should be trivial to review.

PR #193, opened 2017-08-02
Adding backup-restore opsfile using github release

edwardecook:
When the release is uploaded to bosh.io this will be updated to use that instead of the github release url.


staylor14:
Hi @edwardecook!  We accepted the pull request with a couple of caveats.  We've changed the name to deploy-bosh-backup-restore.yml, and we put it on the develop branch.

Thanks much!
steve && @selzoc

PR #194, opened 2017-08-02
Ops file for changing metron agent deployment

jsievers:
this is useful if you have one monitoring stack per
CF deployment and you want to construct the same metric names
based on the deployment name (e.g. "cf" instead of ((system_domain)) ) .


dsabeti:
Hey @jsievers. FYI I've updated the PR to be based on `develop` rather than `master`.

Also, this looks like a pretty useful ops-file, and it could potentially make sense to support this as a first-class feature of cf-deployment. How would you feel if (after review of the PR) we moved this ops-file out of the `community` section and into the main `operations` directory?

jsievers:
sure, we are fine with moving it to the operations directory as officially supported.
I think it's useful as long as there is no [support for wildcards in ops files](https://github.com/cppforlife/go-patch/pull/10), otherwise the change would be only one replace operation.

dsabeti:
Hey @jsievers, I wanted to give you an update about this, because some things have changed. We recently [stopped setting the deployment name explicitly in the manifest](https://github.com/cloudfoundry/cf-deployment/commit/d47ba2c994e486ff02fedf0f8507810e41d600a6), because the metron_agent will use the [deployment name from bosh by default](https://github.com/cloudfoundry/loggregator/blob/develop/jobs/metron_agent/templates/metron_agent.json.erb#L6).

As a result, the way one now renames their metron_agent deployment tag is to rename their bosh deployment. There are a few good reasons to do this -- most importantly that bosh itself tags metrics with the deployment name, so using the same deployment name allows operators to see bosh and cf metrics side by side. However, I wanted to check in and see how this affects your workflow. Do you ever want your metron deployment name to differ from your bosh deployment name?

jsievers:
for our workflow, we always want the metron deployment name to be the same as the bosh deployment name, so the new behaviour is what we want.
Do we want to keep the ops file in case someone would want to change the metron agent name for whatever reason?

dsabeti:
I don't think anyone would want their metron deployment to differ from the
bosh deployment name, so we'll probably remove it.

PR #199, opened 2017-08-08
Remove unused variables

jsievers:
otherwise bosh interpolate --var-errs-unused
will complain


jaresty:
Hey @jsievers we merged your PR and updated the ops file to remove the new router tls_pem property instead of ssl_cert and ssl_key.

jsievers:
@jaresty thanks!

PR #203, opened 2017-08-10
Add cf networking to backup restore VM

jaydunk:
[#149612273]
[#149706241]
[#149612281]

Signed-off-by: Joachim Valdez <valdezj@us.ibm.com>

cc @joachimvaldez

### DIFF
--- a/operations/experimental/deploy-bosh-backup-restore.yml
+++ b/operations/experimental/deploy-bosh-backup-restore.yml
@@ -22,3 +22,11 @@
       jobs:
       - name: database-backup-restorer
         release: backup-and-restore-sdk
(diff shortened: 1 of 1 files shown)


jaresty:
Hey @jaydunk @joachimvaldez (cc: @staylor14) We had to revert this PR since the current cf-networking release (v1.4.0) does not yet contain the `bbr-cfnetworkingdb` job and so the OPs file could not be applied to the current `cf-deployment.yml` and deployed successfully.  Please remind us when you cut a release of `cf-networking-release` with the required job so we can revert our revert.  Thank you!

dsabeti:
Hey @jaydunk, could you explain why we need to  colocate consul with this instance group?

PR #204, opened 2017-08-11
Update Postgres version in bosh-lite opsfile

tcdowney:
Make the bosh-lite Postgres version match the [use-postgres.yml](https://github.com/cloudfoundry/cf-deployment/blob/develop/operations/use-postgres.yml) opsfile.

We would like to use this version since it supports SSL.

Thanks!
@tcdowney && @jenspinney

### DIFF
--- a/operations/bosh-lite.yml
+++ b/operations/bosh-lite.yml
@@ -11,9 +11,9 @@
   path: /releases/-
   value:
     name: postgres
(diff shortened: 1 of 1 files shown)


staylor14:
Hey @tcdowney and @jenspinney!  We went straight to v20 from v16.  If that's a problem for y'all, please re-open.

Thanks much!
--Steve

PR #217, opened 2017-08-18
Update tls_pem to support new format

Nino-K:
- Now cert_chain and private_key must be specified when providing tls_pem
- This PR must be bumped with corresponding change in routing release

[#150267695]

Signed-off-by: Edwin Xie <exie@pivotal.io>

PR #218, opened 2017-08-22
Split router.tls_pem into two properties

aaronshurley:
- Instead of a single string, we are separating tls_pem to include cert_chain
  and private_key.

[#150451369]

Signed-off-by: Gabriel Rosenhouse <grosenhouse@pivotal.io>


aaronshurley:
Didn't realize we already had this one open https://github.com/cloudfoundry/cf-deployment/pull/217

Closing.

PR #219, opened 2017-08-24
Update path from silk-cni to cni

chinangela:
This should not be merged until v1.6.0 is cut. We will bump this thread when that happens.

[#150179537]


chinangela:
Hi,

We have cut v1.6.0, so this PR should be good to merge.

Cheers,
Angela

PR #221, opened 2017-08-29
Fix vm type for bits-service

egurnov:
Since the last pull-request cloudfoundry/cf-deployment#206 was filed, the types of VMs used in deployment changed, which breaks bits-service operations file. This fix makes it use the new VM type.


staylor14:
Hi @egurnov!  We fixed this by hand earlier.  Thanks for the PR, though!

PR #227, opened 2017-09-06
Add ops files for nfs...

julian-hj:
plus incremental nfs/ldap integration, and bbr script deployment for nfs-broker [#150832541](https://www.pivotaltracker.com/story/show/150832541)

NOTE: we test these ops files in the persi pipeline using persi acceptance tests in various flavors, but it might be a good idea to do some smoke level testing of nfs in one of the rel-int pipelines.  If that's something y'all are interested in, let us know and maybe we can XTP when you pull in this PR.

find us on #persi in cf-slack

Thanks!

pretty please

### DIFF
--- a/operations/enable-nfs-volume-service.yml
+++ b/operations/enable-nfs-volume-service.yml
@@ -1,19 +1,18 @@
----
 - type: replace
   path: /releases/-
   value:
     name: nfs-volume
-    version: 1.0.4
-    url: https://bosh.io/d/github.com/cloudfoundry-incubator/nfs-volume-release?v=1.0.4
-    sha1: b2c22394d30c212a9e6dfcdd0abf5879e931ebfe
+    version: 1.0.9
+    url: https://bosh.io/d/github.com/cloudfoundry-incubator/nfs-volume-release?v=1.0.9
+    sha1: 5def43a385b66a8daeee7ca63b18fb7ef9e6b63f
 
 - type: replace
   path: /releases/-
   value:
     name: broker-registrar
-    version: 3.2.2
-    url: https://bosh.io/d/github.com/cloudfoundry-community/broker-registrar-boshrelease?v=3.2.2
-    sha1: f82c6a346d871ccb9835b8e6341f966cf7ebc7c9
+    version: 3.3.1
+    url: https://bosh.io/d/github.com/cloudfoundry-community/broker-registrar-boshrelease?v=3.3.1
+    sha1: 3625828feca7134fd895adf0470bfd91b628bc22
 
 - type: replace
   path: /variables/-
@@ -50,16 +49,18 @@
   path: /instance_groups/-
   value:
     name: nfsbroker
-    azs: [z1, z2]
-    instances: 2
+    azs: [z1]
+    instances: 1
     stemcell: default
-    vm_type: minimal
+    vm_type: m3.medium
     networks: [ name: default ]
     jobs:
     - name: consul_agent
       release: consul
       consumes:
-        consul: {from: consul_server}
+        consul_common: {from: consul_common_link}
+        consul_server: nil
+        consul_client: {from: consul_client_link}
     - name: nfsbroker
       release: nfs-volume
       properties:
@@ -93,7 +94,7 @@
     instances: 1
     azs: [z1]
     lifecycle: errand
-    vm_type: minimal
+    vm_type: default
     stemcell: default
     networks: [{name: default}]
     jobs:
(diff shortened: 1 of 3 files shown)


dsabeti:
Hey @julian-hj. Thanks for the PR! I've got a few pieces of early feedback:
- One of our goals for cf-deployment is to reduce the amount of required configuration that operators will need to provide when they deploy. Part of that means using defaults wherever it's reasonable. I'm going to leave the specific feedback in in-line comments, but if we can use defaults specified by releases, or hard-code them in the manifest, rather than create variables for them, then the operator has less to configure. I'll leave more specific feedback in in-line comments.
- We should probably move all of these files -- including `operations/community/enable-volume-service-nfs.yml` into `operations/experimental` directory
- This PR should be made against the `develop` branch

operations/enable-nfs-volume-service.yml
```diff
@@ -93,7 +94,7 @@
     instances: 1
     azs: [z1]
     lifecycle: errand
-    vm_type: minimal
+    vm_type: default
```
dsabeti: We should probably keep the `vm_type: minimal`. It doesn't really make sense to call a vm_type "default"

operations/experimental/enable-nfs-broker-backup.yml
```diff
@@ -0,0 +1,14 @@
+---
+  - type: replace
+    path: /instance_groups/name=backup-restore/jobs/-
+    value:
+      name: nfsbroker-bbr
+      release: nfs-volume
+      properties:
+        nfsbroker:
+          db_driver: ((nfsbroker-db-driver))
```
dsabeti: `db_driver` is always set to `mysql` in the `enable-volume-service-nfs.yml` ops-file. Let's assume mysql here as well, and we can add a separate ops-file for using postgres if we want.

operations/experimental/enable-nfs-broker-backup.yml
```diff
@@ -0,0 +1,14 @@
...
+          db_username: ((nfsbroker-db-username))
```
dsabeti: This is currently set to `nfs-broker` in `enable-volume-service-nfs.yml`: https://github.com/cloudfoundry/cf-deployment/blob/master/operations/community/enable-volume-service-nfs.yml#L71

operations/experimental/enable-nfs-broker-backup.yml
```diff
@@ -0,0 +1,14 @@
...
+          db_password: ((nfsbroker-db-password))
+          db_hostname: ((nfsbroker-db-hostname))
```
dsabeti: This value is hard-coded as well: https://github.com/cloudfoundry/cf-deployment/blob/master/operations/community/enable-volume-service-nfs.yml#L73

operations/experimental/enable-nfs-broker-backup.yml
```diff
@@ -0,0 +1,14 @@
...
+          db_port: ((nfsbroker-db-port))
+          db_name: ((nfsbroker-db-name))
```
dsabeti: ... as are port and db name: https://github.com/cloudfoundry/cf-deployment/blob/master/operations/community/enable-volume-service-nfs.yml#L74-L75

julian-hj:
Thanks for the feedback.  Fair points all.

We separated out parameters for the bbr database because of the way we run DRATS tests--for DRATS we cf-push the broker and feed it an AWS RDB instance so that we can easily tamper with it during the "disaster" part of the test.  But on reflection, that doesn't really make sense in the context of cf-deployment, and we would need to change it before you would be able to run our DRATS test in your pipelines anyway, so we'll go fix DRATS to use the database and broker that are provisioned in the ops file & resubmit.

Regarding use of the "experimental" folder...are you suggesting that because the feature isn't tested in rel-int pipelines yet, or for some other reason?  nfs-volume-release itself isn't in any way experimental, but if the intent is to put the ops files there until nfs & its tests can be added to a pipeline, then that makes sense to me I guess.

Regarding the PR to master branch...doh.  I intended to PR to develop, but subsequently totally forgot about it.  Sorry!

dsabeti:
@julian-hj, yup, I think I was mistaken to suggest experimental. Let's put all of the ops-files -- including the one that currently lives in the `community` directory -- into the top level `operations` directory. Part of merging this PR (or perhaps a separate story that blocks this PR) will be to add testing in the RelInt pipeline for this.

julian-hj:
closing this PR as it targets the wrong branch--will resubmit improved OPs files to the develop branch early next week.

PR #229, opened 2017-09-11
Remove unused cc properties from worker and clock

tcdowney:
We've recently completed some work on CAPI ([#136759317](https://www.pivotaltracker.com/story/show/136759317)) to remove unused properties from our component spec files.  This made the following properties in the various Cloud Controller jobs unnecessary.

🚫  **WARNING** 🚫 
Please wait until the CAPI release that contains the necessary spec file changes has made it into `cf-deployment` before merging this.

We expect this will be in `capi-release` **1.42.0**, but please double check with us first.  Thanks!

### DIFF
--- a/operations/bosh-lite.yml
+++ b/operations/bosh-lite.yml
@@ -166,34 +166,9 @@
   - public_networks
   - dns
   - load_balancer
-- type: replace
-  path: /instance_groups/name=cc-worker/jobs/name=cloud_controller_worker/properties/cc/default_running_security_groups
-  value:
-  - public_networks
-  - dns
-  - load_balancer
-- type: replace
-  path: /instance_groups/name=cc-clock/jobs/name=cloud_controller_clock/properties/cc/default_running_security_groups
-  value:
-  - public_networks
-  - dns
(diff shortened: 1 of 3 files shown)


operations/bosh-lite.yml
```diff
@@ -166,39 +166,14 @@
   - public_networks
   - dns
   - load_balancer
-- type: replace
```
tcdowney: This was the only ops-file that we use that broke due to this change, but we did not test all of them.

dsabeti:
Hey @tcdowney, it looks like the failure is a result [these operations in `operations/test/add-persistent-isolation-segment-router.yml`](https://github.com/cloudfoundry/cf-deployment/blob/master/operations/test/add-persistent-isolation-segment-router.yml#L91-L97). 

The ops-file is adding the isolation segment's app_domain to the list of app_domains, which are no longer specified for the worker or clock. I'm pretty sure the fix is to update the ops-file not to modify the app_domains for those components.

tcdowney:
Thanks @dsabeti! Those lines have been removed. 🚫

PR #236, opened 2017-09-19
vm_extensions required by cf on bosh-lite

spikymonkey:
Hi cf-deployment,

Following the [instructions in the README](https://github.com/cloudfoundry/cf-deployment#step-1-pave-your-iaas-and-get-a-bosh-director) for using cf-deployment on bosh-lite, I found that the provided cloud-config.yml is missing a couple of required vm_extensions. Hope this helps.

### DIFF
--- a/bosh-lite/cloud-config.yml
+++ b/bosh-lite/cloud-config.yml
@@ -45,6 +45,8 @@ vm_extensions:
     - host: 80
     - host: 443
     - host: 2222
+- name: cf-router-network-properties
+- name: diego-ssh-proxy-network-properties
 - name: cf-tcp-router-network-properties
   cloud_properties:
     ports:


dsabeti:
Hey @spikymonkey. Thanks for the PR. This seems like the correct direction to head in, but there's a bit more to do.

If you look in the bosh-lite ops-file, [we clobber the vm_extensions and add a new one called `ssh-proxy-and-router-lb`](https://github.com/cloudfoundry/cf-deployment/blob/master/operations/bosh-lite.yml#L208-L211). That vm_extension is defined  in the bosh-lite cloud-config, and it opens up [ports for HTTP, HTTPS, and SSH](https://github.com/cloudfoundry/cf-deployment/blob/master/bosh-lite/cloud-config.yml#L42-L47).

To take your PR, we'd probably want to make it so that the two vm_extensions that you've (re-)added each open their necessary ports. Then, we'd want to remove the vm_extension override from the bosh-lite ops-file. After those changes, we'd essentially replace `ssh-proxy-and-router-lb` with two vm_extensions that match the main manifest.

Would y'all be willing to make that change?

spikymonkey:
Hi @dsabeti, thanks for taking a look at this. I've just created #242 as a replacement, so I'll close this one down.

PR #237, opened 2017-09-19
Add UAA client "cc_service_key_client" for CC to talk to credhub

matt-royal:
[#151111701](https://www.pivotaltracker.com/story/show/151111701)

Matt & @tcdowney

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -454,15 +454,19 @@ instance_groups:
               key-1:
                 signingKey: "((uaa_jwt_signing_key.private_key))"
         clients:
+          cc_routing:
+            authorities: routing.router_groups.read
+            authorized-grant-types: client_credentials
+            secret: "((uaa_clients_cc-routing_secret))"
           cc-service-dashboards:
             authorities: clients.read,clients.write,clients.admin
             authorized-grant-types: client_credentials
             scope: openid,cloud_controller_service_permissions.read
             secret: "((uaa_clients_cc-service-dashboards_secret))"
-          cc_routing:
-            authorities: routing.router_groups.read
+          cc_service_key_client:
+            authorities: credhub.read,credhub.write
             authorized-grant-types: client_credentials
-            secret: "((uaa_clients_cc-routing_secret))"
+            secret: "((uaa_clients_cc_service_key_client_secret))"
           cf:
             access-token-validity: 600
             authorities: uaa.none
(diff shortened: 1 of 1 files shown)


dsabeti:
Hey @matt-royal. Would it make sense to add the new UAA client into an ops-file instead of the main manifest? cf-deployment doesn't ship with credhub support just yet, so I think I'd prefer for the UAA client to be optional until then.

tcdowney:
@matt-royal and I addressed @dsabeti's concerns [in Slack](https://cloudfoundry.slack.com/archives/C0FAEKGUQ/p1505845876000235).

The reason why this isn't going into an ops-file is because Cloud Controller is always going to expect to have this UAA client available and will be sending it along to service brokers during the `cf create-service-key` flow. Non-CredHub-aware brokers will just ignore it.

cf-deployment.yml
```diff
@@ -454,15 +454,19 @@ instance_groups:
...
+            secret: "((uaa_clients_cc-routing_secret))"
           cc-service-dashboards:
             authorities: clients.read,clients.write,clients.admin
             authorized-grant-types: client_credentials
             scope: openid,cloud_controller_service_permissions.read
             secret: "((uaa_clients_cc-service-dashboards_secret))"
-          cc_routing:
-            authorities: routing.router_groups.read
+          cc_service_key_client:
+            authorities: credhub.read,credhub.write
```
ishustava: I believe this line should be removed, as cf CredHub has ACLs enabled:
https://github.com/cloudfoundry/cf-deployment/blob/5ef5b808ad82c8251d9b093f6dce3df8b456b79f/operations/experimental/secure-service-credentials.yml#L37-L39
and authorities/scope sections of the UAA token is completely ignored by CredHub.

CredHub for BOSH uses those because it has ACLs disabled. You might have seen it in `bosh-deployment` manifests, but it's not applicable here and could be confusing to the operators.

I'm not entirely sure whether you should completely remove the line, or say something like `authorities: uaa.none`. It should probably be tested out first.

/cc @dsabeti @matt-royal @tcdowney

dsabeti: Thanks @ishustava!

tcdowney: Good to know, that was a misunderstanding on our part and we'll get our story updated. 

Thanks @ishustava!

Looks like these are the [relevant docs](https://github.com/cloudfoundry-incubator/credhub/blob/master/docs/authorization-phase1.md#supported-actors-aka-authenticated-identities) and as long as a credential in CredHub has `uaa-client: cc_service_key_client` we should be able to access it?

ishustava: @tcdowney 

We have tested it, and it looks like we still do require authorities. Sorry for the confusion!

/cc @dannysullivan

tcdowney: Cool, thanks!

PR #242, opened 2017-09-20
Fix replacement of vm_extensions on bosh lite

spikymonkey:
Hi @dsabeti, thanks for taking a look at my last PR (#236). Totally see what you're saying - it looks like I'd missed the bosh-lite ops file completely. I was also looking at your master branch instead of develop - seems like things are almost fixed on develop. I just found one small typo in the bosh-lite ops file that prevents the deployment from proceeding - it results in an extra leading `-` ahead of the `diego-ssh-proxy-network-properties` vm_extension. This had me scratching my head for a while since the error message you get back from bosh is to the effect that 'diego-ssh-proxy-network-properties was not found in the cloud-config', which is confusing since it's clearly there. Using `bosh interpolate` to dump out the generated manifest made it clear what was going on though. Anyway, here's the fix. I'll close #236 since it's redundant now.

### DIFF
--- a/operations/bosh-lite.yml
+++ b/operations/bosh-lite.yml
@@ -207,8 +207,7 @@
   path: /instance_groups/name=diego-brain/jobs/name=ssh_proxy
 - type: replace
   path: /instance_groups/name=router/vm_extensions/-
-  value:
-  - diego-ssh-proxy-network-properties
+  value: diego-ssh-proxy-network-properties
 - type: replace
   path: /instance_groups/name=router/jobs/-
   value:


dsabeti:
Hey @spikymonkey. We merged the PR earlier today, but, due to another story we were working on, we discovered that stacking vm_extensions is a tricky business and had to recombine the the vm_extensions in this commit: https://github.com/cloudfoundry/cf-deployment/commit/a17b4eb0cd2b826201741a0c9fd74fd776409304.

We're still working on trying to unify the vm_extension name between a normal cf-deployment and cf-deployment on bosh-lite. Were there any other concerns you were trying to address when you submitted the first PR?

spikymonkey:
Hi @dsabeti, no not really - just trying to get `cf-deployment` working on fresh bosh-lite out of the box.

PR #243, opened 2017-09-20
Fix router vm extensions in bosh-lite

egurnov:
`/-` in path appends to array, the value should be scalar.


spikymonkey:
Ha, we both found the same thing :) #242

dsabeti:
Good catch @egurnov. I merged @spikymonkey's PR (just the order I found them in), which is pretty much identical to this one.

PR #246, opened 2017-09-20
Add cf syslog drain

bradylove:
- Adds cf-syslog-drain-release
- Adds reverse_log_proxy from Loggregator
- Removes etcd
- Removes syslog_drain_binder


dsabeti:
Hey @bradylove. Thanks for this PR! We're really excited to get etcd out of cf-deployment. One thing: we'll need to ship two versions of cf-deployment, the first that introduces cf-syslog-drain-release but doesn't remove etcd, and then a subsequent one that removes etcd. It has to do with with rolling deploys. If we try to update everything at once, then etcd will get removed from the deployment before the dopplers get updated with changes to stop using etcd.

Could we have one PR that replaces the syslog-drain-binder with cf-syslog-drain-release (and reverse log proxy), and then a separate PR that removes etcd?

bradylove:
@dsabeti Sure. I will close this PR and re-open with two separate PRs.

PR #253, opened 2017-09-21
Add openstack support

voelzmo:


openstack/README.md
```diff
@@ -0,0 +1,103 @@
...
+bosh -d cf deploy ../cf-deployment.yml \
+  -o ../operations/openstack/use-compiled-releases.yml \
+  -o ../operations/openstack/ephemeral-disk-vm-extensions.yml \
+  --vars-store ~/deployments/cf/cf-vars-store.yml \
+  -v system_domain="<system-domain>"
+```
+
+### With TCP Routing
+
+If you want to enable TCP Routing, the Terraform module should have run with `use_tcp_routing=true`. Add the following line to the deploy cmd:
```
dsabeti: @voelzmo I think you might have missed this, but we recently inlined TCP Routing from the ops-file into cf-deployment.yml. I think we can just get rid of this section  -- assuming the terraform template creates TCP LBs by default, which seems to be the case.

operations/openstack/ephemeral-disk-vm-extensions.yml
```diff
@@ -0,0 +1,14 @@
+---
```
dsabeti: Since this is the only _required_ ops-file for openstack, would it make sense to just call this `operations/openstack.yml`? That's currently our approach to other IaaS-specific ops-files.

operations/openstack/use-swift-blobstore.yml
```diff
@@ -0,0 +1,151 @@
+---
```
dsabeti: Even though this is ops-file is IaaS-specific, it's not required for deploying to Openstack. Maybe just move this one directory up -- `operations/use-swift-blobstore.yml`?

openstack/README.md
voelzmo: Nice, we did miss that, thanks!

operations/openstack/ephemeral-disk-vm-extensions.yml
voelzmo: I'm not sure about that. I like the extra transparency of what this actually does. But sure, we can make this go in line with the other IaaS-specific ops files.

operations/openstack/use-swift-blobstore.yml
voelzmo: You are correct. In theory I could imagine people installing *just swift* to get an on-premise blobstore.

voelzmo:
@dsabeti Sure, I think the `openstack` directory is only a temporary solution for as long as installation on OpenStack is different than for the IaaS layers where `bbl` can be used. Once that is streamlined, we should get rid of it.

selzoc:
@voelzmo we merged this with some modifications.  please submit another PR if you want to re-arrange some more.  Most notably, we changed the directory structure and renamed the flavors.

PR #255, opened 2017-09-22
Bypass cc-bridge by default

matt-royal:
By default, disable the cc-bridge and have CC and Diego talk with mTLS

https://www.pivotaltracker.com/story/show/150421205

Thanks,
Matt && @rainmaker


dsabeti:
Hey @matt-royal @rainmaker, thanks for the PR! This mostly looks good, but I was wondering if y'all had any opposition to not allowing operators to enable the CC Bridge -- essentially, instead of making Bridge Consumption the default configuration, could we make it the _only_ configuration? Maybe @zrob has thoughts on this?

If so, then instead of renaming `bypass-cc-bridge.yml` to `enable-cc-bridge.yml`, we could keep the name and just empty the ops-file out.

matt-royal:
We made the requested change after confirming with @zrob

PR #265, opened 2017-10-02
OpenStack's cloud-config - missing disk type: 1GB

wildloop:
in reference to the disk definition for syslog-scheduler: https://github.com/cloudfoundry/cf-deployment/blob/master/cf-deployment.yml#L1293


dsabeti:
Hi @wildloop, thanks for catching that. It turns out the `1GB` persistent disk is not actually necessary for that job, so I have a story to remove it:
https://www.pivotaltracker.com/story/show/151608531

Does that address your concerns? If so, I think we can close the PR.

wildloop:
Yes, thanks, it's ok, I just wanted to workaround the deploy error.

PR #268, opened 2017-10-04
Add capi job and set release_level_backup flag to true in backup-restore opsfile

edwardecook:
The CAPI job needs to be added and the release_level_backup flag set to true in order for the cloud controller data to actually be backup up.


dsabeti:
Hi @edwardecook, could you sign the CLA (as described by @cfdreddbot), and resubmit the PR?

PR #279, opened 2017-10-16
Split CredHub load balancer into separate opsfile

ebeer:
@dsabeti 

https://www.pivotaltracker.com/story/show/151777717

[#151777717] Operators can deploy cf-deployment with CredHub without an LB vm extension by default


tophat8855:
We resolved the conflicts locally and pushed. All is well.

PR #280, opened 2017-10-16
update winc-release version

sunjayBhatia:
update `winc-release` to `0.5.0`


sunjayBhatia:
actually, closing since this is a little premature, some other changes need to be added to the opsfile at the same time as this change

PR #288, opened 2017-10-18
Create an ASG rule for credhub

ljfranklin:
- Since the latest recommended approach is to have apps talk directly to
  credhub's internal IP, CF requires an Application Security Group for
  credhub
- Specifying ASGs in the manifest only works on fresh deploys, adding
  this ops file to an existing env will require adding the ASG manually

[#152045213]

Signed-off-by: Lyle Franklin <lfranklin@pivotal.io>

### DIFF
--- a/operations/experimental/secure-service-credentials.yml
+++ b/operations/experimental/secure-service-credentials.yml
@@ -83,6 +83,21 @@
 - type: replace
   path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/credhub_api?/ca_cert
   value: ((credhub_tls.ca))
+# Note: these groups are only created on the first deploy, groups must be created manually for existing envs
+- type: replace
+  path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/cc/security_group_definitions?/-
+  value:
+    name: credhub
+    rules:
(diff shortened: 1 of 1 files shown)


dsabeti:
@ljfranklin Would this potentially make more sense as a Concourse task for updating the ASGs through the CLI? We could host it in [cf-deployment-concourse-tasks](https://github.com/cloudfoundry/cf-deployment-concourse-tasks).

ishustava:
Hey @ljfranklin,

Does adding this to `cf-deployment-concourse-tasks` sound good to you? We have a tracker [story](https://www.pivotaltracker.com/story/show/152164467) to add/update ASGs.

ljfranklin:
@ishustava @dsabeti up to y'all. I'd vote for the ops-file approach as you only need to include the ops-file and everything works. With a task, users would probably deploy, see it fail due to missing ASG, then maybe they discover the shared task or maybe they re-invent the wheel with a new task. If you're more comfortable maintaining a separate task that's also fine.

dsabeti:
@ljfranklin, I think we'll go forward with the task. For me, the thing that's important is that an operator would have to delete their deployment and then redeploy from scratch if we added the ASG definition to the manifest instead.

PR #289, opened 2017-10-18
Add ops file for adding syslog_forwarder

lisamcho:
https://www.pivotaltracker.com/story/show/152053493

Signed-off-by: Greg Cobb <gcobb@pivotal.io>

### DIFF
--- a/operations/use-syslog-addon.yml
+++ b/operations/use-syslog-addon.yml
@@ -1,0 +1,25 @@
+- type: replace
+  path: /addons?/-
+  value:
+    name: syslog_forwarder
+    include:
+      stemcell:
+      - os: ubuntu-trusty
+    jobs:
+    - name: syslog_forwarder
+      release: syslog
+    properties:
+      syslog:
+        address: ((syslog_remote_address))
+        port: ((syslog_remote_port))
+        transport: tcp
+        fallback_servers: []
+        custom_rule: ""
+
+- type: replace
+  path: /releases/-
+  value:
+    name: syslog
+    version: 11
+    url: https://bosh.io/d/github.com/cloudfoundry/syslog-release?v=11
+    sha1: 332ac15609b220a3fdf5efad0e0aa069d8235788


dsabeti:
Hi @lisamcho, were y'all aware of [this runtime-config](https://github.com/cloudfoundry/cf-deployment/blob/master/runtime-configs/enable-component-syslog.yml)? It does exactly the same thing as the ops-file in this PR.

Gerg:
@dsabeti There are a couple differences:
1.  The runtime config is not composable, so we can't use it with other runtime configs.
2. The ops file version can work in a deployment manifest or as a runtime config (we use it in a deployment manifest)
3. The runtime config does not work with windows cells. Our ops file does not work with windows cells either, but we can omit it from windows cells.
4. Our variable names are better

operations/use-syslog-addon.yml
```diff
@@ -0,0 +1,25 @@
...
+    jobs:
+    - name: syslog_forwarder
+      release: syslog
+    properties:
+      syslog:
+        address: ((syslog_remote_address))
+        port: ((syslog_remote_port))
+        transport: tcp
+        fallback_servers: []
+        custom_rule: ""
```
dsabeti: Can we expose `transport`, `fallback_servers`, and `custom_rule` as we do in the runtime-config? We can go with names like the one's you've used (`syslog_transport`, etc.).

staylor14: The new names are superior, but changing the names will require some communication to current users.  We should think about a means of deprecating existing ops files, maybe.

loewenstein:
Some thoughts:
1. runtime-configs are composable, just specify a name, e.g. `bosh urc --name syslog syslog-forwarder.yml`
2. What is the benefit of having it in the deployment manifest?
3. You can use placement rules in runtime-config too (see http://bosh.io/docs/runtime-config.html#placement-rules)
4. Can we have better names in the existing ops file?


cc @cppforlife wdyt?

dsabeti:
@loewenstein Thanks for bringing this stuff up. Also, in light of the fact that placement rules can be made in the runtime-config, runtime-configs being composable, and having [this PR to make the names better in the runtime-config](https://github.com/cloudfoundry/cf-deployment/pull/282), maybe we can just move forward improving the runtime-config, @Gerg?

rainmaker:
Hi friends,

I had a productive conversation today with @dsabeti and @anEXPer about the relationship between runtime-config and ops files in `cf-deployment`. 

To reiterate, CAPI likes the idea of:

- Continuing to use the [bosh-deployment-resource](https://github.com/concourse/bosh-deployment-resource) in our pipeline, while maintaining the ability to browse all the changes we make to cf-deployment from the vanilla setup. Ideally, this would include changes to the runtime config. As far as we know, the resource doesn't currently support updates to the runtime config.
- Being able to apply this particular syslog configuration to a single deployment, using the existing ops-file strategy. This is useful in the context of a test environment, where we'd like to store and read a single deployment's config without worrying about the global state of the director. (For the record, we'd be fine with both a runtime config AND an ops file, but understand that it'd be annoying to maintain 2 versions of the same syslog configuration.)

So, we discussed those items as well as `cf-deployment` concerns around supporting both operator and platform-related configurations. We then realized that it *should* now be possible to offer a single ops file, supplied to either `bosh update-runtime-config` or `bosh deploy` using the `-o` flag. We really liked the idea of maintaining a single file that would both be useful for operators who set their runtime config across an entire director's deployments, as well as developers like CAPI who want to control the configuration of each test deployment individually.

Sounds like the release-integration team will hammer out the details to fully implement this idea. For now, would you prefer not to accept this PR until you decide where to store these multi-purpose ops files and how to communicate about their use? Or, you could accept it and move it to a final location later :)

@Gerg @lisamcho and any other CAPI friends, I had this conversation during your lunchtime so feel free to chime in with any other questions or thoughts from our side.

Thanks!
Raina
CF CAPI team

dsabeti:
@rainmaker, let's close this PR, and we'll start from scratch with this story:
https://www.pivotaltracker.com/story/show/152257058

dsabeti:
I'm going to close this ops-file since we've built the ops-file here: https://github.com/cloudfoundry/cf-deployment/blob/master/operations/addons/enable-component-syslog.yml

PR #300, opened 2017-10-25
Ops file for deploying garden-runc in oci-phase-1

BooleanCat:
Deploying with this operation switches garden to creating images in the OCI compatible format.

[#151724164]


BooleanCat:
Resubmitting.

PR #301, opened 2017-10-25
Ops file for deploying garden-runc in oci-phase-1

BooleanCat:
Deploying with this operation switches garden to creating images in the OCI compatible format.

[#151724164]


staylor14:
@BooleanCat Thanks for the pull request!

Couple of questions regarding your paths.  On
```
- type: replace
  path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/cc?/diego?/temporary_oci_buildpack_mode?
  value: &temporary_oci_buildpack_mode oci-phase-1
```

1) the question marks for `diego` and `temporary_oci_buildpack_mode` are extraneous, since the first question mark (on `cc`) carries over to them.
2) perhaps more importantly, though, you are saying "We're fine if `cc/diego` aren't defined, which I think is probably not the case - if you're applying this opsfile to a manifest without those properties, I *think* it would mean something very bad™ has happened.  I think it may be better to fail the moment bosh can detect this, rather than to allow a deployment to fail.

I'm proposing we change this path to `/instance_groups/name=api/jobs/name=cloud_controller_ng/properties/cc/diego/temporary_oci_buildpack_mode?`  I can make the change if you agree that it's better to fail fast.  If you don't, I can just remove the extraneous question marks and accept it as-is (semantically).

Regards,
--Steve

BooleanCat:
Hi Steve - I agree that failing fast is preferable. Would you mind making that change? I already left the office and it would be great to have this merged as soon as possible.

Thanks!

staylor14:
That's exactly what I was hoping we'd do.  We'll make the update on our side.  It should be ready for you when you get in :)

--Steve && @sphawley

PR #305, opened 2017-10-26
Change CA references to ca-name.certificate instead of certificate.ca

danjahner:
Hey Team - 

The current manifest references certificate authorities as `certificate-name.ca`. This PR changes that to `ca-name.certificate`. 

There are two primary motivations for wanting this change:

1. Clarity on which CA is trusted by which components. 

Example: 
```
  - name: auctioneer
    release: diego
    properties:
      diego:
        auctioneer:
          ca_cert: "((diego_auctioneer_server.ca))"
```
It is unclear that the above job trusts the `service_cf_internal_ca` CA. 

2. Ability to gradually rotate CAs.

The accepted method to rotate a CA without downtime is a multi-step process. The first step adds an additional new CA, next its dependent certificates are rotated and signed by the new CA, and finally the old CA is removed. This process requires the CA references to be decoupled from the certificate, because the desired CA value may not necessarily match to the CA which signed the certificate.

No end-user functionality would be affected by this change.


dsabeti:
Hi @danjahner. Thanks for this PR. It's an interesting change, because we had specific commits that opted for the certificates' `.ca` reference: https://github.com/cloudfoundry/cf-deployment/commit/6b3b9d6d0e0b0fb3a7ddc3e56c1c8c790b322d20. 

At the time, we were getting feedback that, given a certificate X, we should use `X.ca` instead of `X_ca.certificate` because `X.ca` would always be correct --`X_ca.certificate`, on the other hand, could reference a CA that didn't actually sign `X.certificate`, especially if the CA changed:
```
variables:
- name: X
  type: certificate
  options:
    ca: global_CA
- name: global_CA
  type: certificate
  options:
    is_ca: true
- name: X_ca
  type: certificate
  options:
    is_ca: true
```

That feedback was given when people were still getting used the `variables` section of the manifest. The change in CA in those cases weren't about CA rotation, but about changing requirements in mTLS architecture, and operators would have to delete the certs from the vars-store and regenerate them with the new CA.

I'm trying to walk through the process of rotating the two CAs in these two schemes, but compiling thoughts in text is going to be hard. Let's chat in-person to go through this so I can understand.

danjahner:
Hey David - 

Happy to chat in person. For posterity (aka for reference when another PR to flip it back appears 😄 ), I'll add some clarity on the CA rotation bit. 

The current understanding for credential references is that they contain exactly one value. However, for rotation cases, that may not be desirable. Instead, you want the credential to reference all _active_ values.

Example manifest:
```
  - name: auctioneer
    release: diego
    properties:
      diego:
        auctioneer:
          ca_cert: "((service_cf_internal_ca.certificate))"
```

Non rotation case would be templated by BOSH like: 
```
  - name: auctioneer
    release: diego
    properties:
      diego:
        auctioneer:
          ca_cert: |
            -----BEGIN CERTIFICATE-----
            ...1...
            -----END CERTIFICATE-----
```
However, the gradual rotation case could be templated by BOSH like: 
```
  - name: auctioneer
    release: diego
    properties:
      diego:
        auctioneer:
          ca_cert: |
            -----BEGIN CERTIFICATE-----
            ...2...
            -----END CERTIFICATE-----
            -----BEGIN CERTIFICATE-----
            ...1...
            -----END CERTIFICATE-----
```

This model also clarifies the purpose of these fields and highlights the fact that the ca_cert provided has a separate function from the TLS credentials. The CA defines what to trust when making outbound connections, whereas the TLS credentials define what values to use for server connections. Although `certificate.ca` often contains an appropriate value for this other need, that is not always the case.

```
  - name: auctioneer
    release: diego
    properties:
      diego:
        auctioneer:
          ca_cert:        <<certificate(s) to trust when validating an outbound connection to a service
          certificate:   <<certificate to use for server connections
          private_key: <<private key to use for server connections
```

crawsible:
To bump this thread: I agree with @danjahner that it is more semantically correct to declare trust for a CA directly, rather than whichever CA happened to sign some leaf cert. It's also the case that supporting zero-downtime CA rotation without manifest manipulation is a lot easier when CAs are referenced using `ca.certificate`. Ultimately, whether zero-downtime rotation is possible when `certificate.ca` is used is up to BOSH. It would be more work for BOSH to support this, but if they'd like to, then we're happy to assist where necessary.

In order to support zero-downtime rotation without manual manifest manipulation, CredHub has learned how to return either one or two certificate values (including CA certificates) in an array when a user asks for the "current" value at a given location. In the solution @cppforlife and I worked out about a month ago, when two values are returned BOSH will stack the "cert" values of each and place the stack wherever `.certificate` is found.

CredHub is unwilling to return two `ca` values for a single leaf or intermediate certificate, as this is misleading: the certificate PEM has a signatory, and we shouldn't pretend as though it has multiple signatories when it does not.

For BOSH to support zero-downtime rotation with `cert.ca`-referenced CAs, it will need to be smarter than just returning the "ca" value of the returned JSON object for the leaf certificate, and instead make a request to CredHub for the current value of the named CA certificate.

dsabeti:
Thanks for bumping this thread @crawsible. I was supposed to reach out to @danjahner a while ago to discuss this in-person, but I dropped the ball.

As soon as I read 
> declare trust for a CA directly

I realized that my understanding of the problem was backwards. I wanted to put the leaf certificate before the CA certificate -- "I want to validate certificate X, so I need to know which CA signed it" -- when instead I should be first declaring trust for a CA (or set of CAs), and then validate that any certificate I use is signed by something I trust.

This also helps me understand the workflow for CA rotation. First, the manifest will need to declare trust for a second CA -- in which case `ca.certificate` will render multiple CA certs in the templates -- and then validate that any new leaf certificate is signed by one of the trusted CAs.

It looks like there are some merge conflicts with this PR. Does it make sense to cross-team pair on it next week?

danjahner:
I fixed the merge conflict. I tested the CA to cert mapping back in October and it doesn't appear that this has changed since, but you would know better than I would whether changes have been made in the interim. 

Let me know if you have any questions!

dsabeti:
Thanks @danjahner!

aeijdenberg:
@crawsible @cppforlife - I read with great interest the comment from Jan 6th above that states:

>  In the solution @cppforlife and I worked out about a month ago, when two values are returned BOSH will stack the "cert" values of each and place the stack wherever .certificate is found.

I can't find any evidence `bosh` is currently doing this. Are there links to any issues to follow?

(we're considering writing a patch to `bosh-cli` to essentially do this - it would be better if the bosh director already had a mechanism to do so)

aeijdenberg:
@crawsible @cppforlife - as I mentioned in my previous comment, we're quite keen to find a solution for certification rotation. To that end, we've made a small patch to bosh-cli to do so (rotates certs in a 3 pass deployment): https://github.com/cloudfoundry/bosh-cli/pull/490

It sounds like you may have been cooking up a similar or alternate approach - it would be good to compare notes?

crawsible:
Hi @aeijdenberg — I'll 302 you to @freddesbiens for an update on BOSH's work integrating the CredHub versions feature, which was built to enable this workflow.

PR #307, opened 2017-10-30
Rename GrootFS BOSH properties

glestaris:
`grootfs.remote_layer_tls.*` got renamed to `tls.*` in GrootFS [v0.29.0](https://github.com/cloudfoundry/grootfs-release/releases/tag/v0.29.0).

### DIFF
--- a/operations/experimental/enable-oci-phase-1.yml
+++ b/operations/experimental/enable-oci-phase-1.yml
@@ -12,7 +12,7 @@
   value: *temporary_oci_buildpack_mode
 
 - type: replace
-  path: /instance_groups/name=diego-cell/jobs/name=grootfs/properties/grootfs/remote_layer_tls?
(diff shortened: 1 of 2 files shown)


dsabeti:
Hey @glestaris. It looks like the commits from your PR are in the history of the `master` branch, but for some reason Github doesn't seem to think we've merged this PR. (My guess is that we accidentally did a rebase instead of a merge.)

On the current `master` branch, we've [removed `grootfs.remote_layer_tls` in favor of just `tls`](https://github.com/cloudfoundry/cf-deployment/blob/v1.2.0/operations/experimental/enable-oci-phase-1.yml), and stripped out the [deprecated properties](https://github.com/cloudfoundry/cf-deployment/blob/v1.2.0/operations/experimental/use-grootfs.yml#L1-L5). I'll close this PR, but feel free to re-open if it doesn't look right.

cc @julz so that context is shared with current garden team members.

PR #319, opened 2017-11-03
Stop enabling `bpm` on uaa instances

aramprice:


aramprice:
This PR removes the `bpm.enabled` flag from UAA instances because the current version of `uaa-release` is not yet compatible with `bpm-release` version `0.2.0`.

Once a compatible version of UAA has been releases we expect to revert this change.

PR #320, opened 2017-11-03
Stop enabling `bpm` on CAPI instances

aramprice:


aramprice:
This PR removes the `bpm.enabled` flag from CAPI instances because the current version of `capi-release` is not yet compatible with `bpm-release` version `0.2.0`.

Once a compatible version of CAPI has been releases we expect to revert this change.

PR #322, opened 2017-11-06
explicit 'bosh' cf/uaa admin user; so auditors look in bosh tasks

drnic:
From conversations with @zrob we would like bosh deploy/bosh run-errand
to have its own user so as to help CF auditors know who/where to look
for who did what.

With this PR, running `bosh run-errand smoke-tests` will generate
an event with `bosh` as the user:

    cf curl /v2/events\?q=type:audit.app.delete-request
    ...
         "entity": {
            "type": "audit.app.delete-request",
            "actor": "ec5a6b63-27ba-4f3c-a33b-896d6b358dd7",
            "actor_type": "user",
            "actor_name": "bosh",
            "actor_username": "bosh",

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -45,8 +45,8 @@ instance_groups:
       smoke_tests:
         api: "https://api.((system_domain))"
         apps_domain: "((system_domain))"
-        user: admin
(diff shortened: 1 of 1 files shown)


dsabeti:
👍

PR #324, opened 2017-11-07
Separate out cf-syslog-drain adapters by default

JohannaSmith:
The cf-syslog-drain `adapters` should not be on the `doppler` vm.

[#152519481](https://www.pivotaltracker.com/story/show/152519481)

/cc @apoydence


dsabeti:
Hey @JohannaSmith. Could you also update the `scale-to-one-az.yml` ops-file to lower the number of adapter vms?

dsabeti:
Also @JohannaSmith, we just realized that this change is going to cause syslog downtime. Currently, the adapters are colocated with the dopplers, which get updated well before the new adapters get deployed. That means that for some amount of time, there will be no adapters whatsoever.

I think the best way to handle this is to hold off on merging this, ship a version of cf-deployment warning everyone that they should deploy with the `non-collocated-cf-syslog-drain.yml` ops-file in order to avoid downtime in the future. Then, later, we merge the PR and ship a new version.

PR #331, opened 2017-11-15
Update README.md for bosh-lite

apoydence:
Update the vars-store filename to match the name given in the .gitignore.


apoydence:
I used master as the base at first, but I have since updated it.

PR #334, opened 2017-11-20
Fix: scale resource_group "adapter" to one az

petergtz:
Fixes commit 6b817d1012d522ea6fe860234a5bc7b9cb1a2915.

### DIFF
--- a/operations/scale-to-one-az.yml
+++ b/operations/scale-to-one-az.yml
@@ -76,6 +76,9 @@
   path: /instance_groups/name=cc-worker/azs
   value: [ z1 ]
 - type: replace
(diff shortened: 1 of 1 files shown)


ben-schumacher:
@selzoc Will that patch eventually make it into cloudfoundry:master?  https://github.com/cloudfoundry/cf-deployment/pull/334/commits/b2feb0aaa047c25254eb4b8622b5c3cb86f0716c

dsabeti:
@ben-schumacher, the change needs to make it's way through our testing pipelines. Then, yes, it will be available on the master branch. I typically cut new versions of cf-deployment twice per week.

PR #342, opened 2017-11-27
Reduce UAA logging level from DEBUG to INFO

anEXPer:
The default UAA logging level is DEBUG. This is not an appropriate default for Cloud Foundry deployments. We'd like to set it to INFO, the level at which access log information is emitted, in cf-d.

We spoke with the UAA team, and they did not have a compelling reason for the current log level. In a default-scale deployment of cf-d, UAA DEBUG logs account for more than 40% of total component log volume.

We are following up with UAA about reducing the default in their release, at which point this additional configuration could be removed.


anEXPer:
It looks like time has marched on a bit since we made/tested this. We'll get develop merged and the tests passing shortly.

PR #354, opened 2017-12-14
Add workaround for Diego memory calculation bug

emalm:
From the amended [Diego v1.31.0 release notes](https://github.com/cloudfoundry/diego-release/releases/tag/v1.31.0):

> We have identified an issue with the calculation of the memory container metrics in this version of Diego for those cells that do not enable the container proxy (which is the default configuration). For apps with a memory limit under 18 MB, the miscalculation results in reporting a very large number for the container memory usage, which causes `cf push` to fail when checking on the app instance status with an error of the form `Could not fetch instance count: Invalid JSON response from server: json: cannot unmarshal number 18446744073708089344 into Go struct field .Mem of type int64`. We are currently working to address this issue in [story #153655210](https://www.pivotaltracker.com/story/show/153655210) and will include the fix in a new final Diego release as soon as possible.

> As a workaround, operators should set the `containers.proxy.additional_memory_allocation_mb` property to `0` on the `rep` and `rep_windows` jobs on their Diego cells.

This workaround should be reverted once a new final Diego version exists with a fix for this issue and is included in cf-deployment.

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -1094,6 +1094,9 @@ instance_groups:
   - name: rep
     release: diego
     properties:
+      containers:
+        proxy:
+          additional_memory_allocation_mb: 0
       diego:
         rep:
           bbs: *diego_bbs_client_properties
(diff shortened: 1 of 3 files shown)


emalm:
Closing in favor of bumping to Diego v1.31.1 in https://github.com/cloudfoundry/cf-deployment/commit/1179c1b37c8de04d70490902c28c6f0c76c17f44.

PR #355, opened 2017-12-14
Update defaults to use make use of Groot in Garden

williammartin:
Groot is the default image manager in garden-runc 1.11.0 (not-yet-released) and requires
tar rootfs paths. Provide a legacy shed ops file so that deployments can
temporarily use garden-shed during deprecation period.


williammartin:
PRed against the wrong branch, closing.

PR #356, opened 2017-12-14
Update defaults to use make use of Groot in Garden

williammartin:
Groot is the default image manager in garden-runc 1.11.0 (not yet released) and requires
tar rootfs paths. Provide a legacy shed ops file so that deployments can
temporarily use garden-shed during deprecation period.


williammartin:
Hey cf-deploymenters,

n the next release of Garden (coming soon), `GrootFS` is replacing `garden-shed` as the default image management piece and this requires a few changes in cf-deployment:

There was an experimental GrootFS opsfile which we can now get rid of in favour of a legacy shed ops file. I wasn't sure where to put this new ops file since the legacy folder readme suggests it is only for `cf-release` artifacts. 

Now that GrootFS is the default, we don't need to use the groot ops file for rootless-containers or oci-phase1 ops files. 

Finally, I updated cf-deployment.yml to use `.tar` rootfs by default (with shed reverting them to directories), in the same way that the GrootFS opsfile used to do.

With that said, need some advice on this PR. Since we're pulling the GrootFS job into Garden, anyone using the experimental groot ops file, deploying with the next release of garden-runc will run into BOSH errors like:

```
Task 41 | 14:33:26 | Error: Package name collision detected in instance group 'diego-cell': job 'garden-runc/garden' depends on package 'garden-runc/grootfs', job 'grootfs/grootfs' depends on 'grootfs/grootfs'. BOSH cannot currently collocate two packages with identical names from separate releases.
```

How are these breaking changes normally communicated?

Anyone who doesn't add the shed ops file will start using Groot (which is fine as an upgrade path). It is also fine to go from garden-runc 1.10+cf-deployment/master+groot.ops to garden-runc 1.11+cf-deployment/thisPR.

Thanks,

Will

Also CC @julz

julz:
Just a little extra note: although it's safe to upgrade to the new garden-with-groot from any release of garden with or without the extra (and now removed) groot ops file we would recommend that folks try to time the upgrade with a stemcell update if possible (or else do a recreate). If not there may be some leftover files which'll take up disk space, depending on whether evacuation completes or not. We're going to make a note about this on our release notes for the next garden version, but it may make sense to have a similar note on the cf-deployment release notes for the version this PR makes it in to?

williammartin:
@julz Even evacuation won't clear up unused layers in the graph, only the RW layers.

julz:
yeah, that was a super unclear sentence: I meant to say the _amount_ of space taken will depend on whether evacuation completes or not, but yeah you're right: to be clear there'll definitely be _some_ shared layers that won't get collected till the next stemcell or recreate, right?

dsabeti:
Hey @williammartin and @julz. We're super excited for GrootFS to be promoted to the default filesystem.

Let me make sure I understand everything that's happening here:
1. GrootFS will be inlined into garden-runc-release. One result of this is that people who already use GrootFS via grootfs-release will run into a name collision for the packages.
2. Upgrading may leak some files (that are ephemeral?), and a `bosh recreate` or stemcell roll would clear that up.

We typically communicate breaking changes by making a major version bump, but we still need an upgrade path for people who are currently using GrootFS. Is there currently any way to upgrade a Diego cell that uses GrootFS from grootfs-release to GrootFS from garden-runc-release?

dsabeti:
cc @staylor14 @selzoc

williammartin:
@dsabeti 

1: Yes, it has been inlined and will be part of the 1.11.0 release. Using that and the `groot.ops.yml` that currently exists will result in name collision.
2: Correct, in that `garden-shed` stores images in a different location than `grootfs`, and this store is on the ephemeral disk.

Unless I misunderstand your question, someone deploying CF right now via `cf-deployment` using `cf-deployment.yml` + `groot.ops.yml` should simply be able to stop using `groot.ops.yml`, when they bump `garden-runc-release` to 1.11.0.

dsabeti:
@williammartin, ok, I understand now. The name collision only occurs if you use the garden-runc-release 1.11.0 _and_ the grootfs ops-file.

The easiest way around this is to keep the grootfs ops-file but empty its contents in order to render it a no-op. [We have plenty of examples for this already](https://github.com/cloudfoundry/cf-deployment/blob/master/operations/experimental/disable-consul-service-registrations-locket.yml). We'll communicate clearly in release notes that the ops-file no longer does anything, and when we cut cf-deployment 2.0, we'll remove the old ops-file.

As for the graph cleanup issue, we can also document that operators may want to recreate their Diego cells, but I also wonder if there's a way to clean this up in a pre-start script of something like that? (cc @cppforlife in case he has any ideas)

Callisto13:
@dsabeti we've pushed a noop use-grootfs. I think @julz wants a chat about migrations

dsabeti:
🎉

PR #358, opened 2017-12-15
Add shim to disable Consul on BOSH-Lite

emalm:
This change alters the BOSH-Lite operations file to tolerate the absence
of the `consul` instance group when scaling down to 1 instance. When
Consul has already been disabled, this results in the recreation of an
(invalid) `consul` instance group, so this also adds a shim operations
file to delete it.


emalm:
cc: @chendrix

PR #369, opened 2018-01-03
Introduce a shutdown delay when running in HA mode

APShirley:
- BOSH DNS polls for health every 20s based on the monit status of an
 instance group. The proxy process should stay healthy for longer than it reports as healthy to
 give it time to be removed from the DNS pool.

Signed-off-by: Nitya Dhanushkodi <ndhanushkodi@pivotal.io>


APShirley:
Just confirming this passed CATS locally (sans v3 tests).

PR #370, opened 2018-01-04
Use gateway for DNS server in windows2016 containers

sesmith177:
Windows server containers now use 172.30.0.1 as their DNS server instead of 169.254.0.2

This is to enable BOSH DNS for containers via PR: https://github.com/cloudfoundry/bosh-dns-release/pull/14


selzoc:
Hi @sesmith177 @aminjam am I to understand that this won't work until that bosh-dns PR has been merged and a release created from that?

sesmith177:
@selzoc That is true. However, the current value in `use-bosh-dns-for-windows2016-containers.yml` has never actually worked, so there's no need to block merging this PR

selzoc:
LOL OK thanks!

PR #371, opened 2018-01-04
spelling upgrade

flavorjones:
I believe they call this an "[obvious fix](https://cla.pivotal.io/about#obvious-fixes)" ;)


dsabeti:
Thanks @flavorjones !

PR #374, opened 2018-01-05
Share doppler link

apoydence:


selzoc:
@apoydence Hi, this looks like it shares a doppler link from the `diego-api` instance group - is that intended, or was this supposed to go on the `doppler` instance group?

apoydence:
@selzoc No that was a mistake from merging. I've fixed it.

selzoc:
Thanks @apoydence unblocked!

selzoc:
Hi @apoydence - upon further reflection, this seems like it's in the wrong place again - are you trying to share the link of the `consul_agent` on the `doppler` instance, or the `doppler` job?  See this for how it's done on the `reverse_log_proxy`: https://github.com/cloudfoundry/cf-deployment/blob/master/cf-deployment.yml#L1210-L1213

apoydence:
@selzoc You're right again! I've moved it to the doppler job (instead of the consul one).

PR #382, opened 2018-01-11
Update windows opsfile release versions

sunjayBhatia:
Updates opsfiles for `windows2016` and `windows2012R2` cells

### DIFF
--- a/operations/experimental/use-offline-windows2016fs.yml
+++ b/operations/experimental/use-offline-windows2016fs.yml
@@ -2,4 +2,4 @@
   type: replace
   value:
     name: windows2016fs
-    version: 0.0.17
+    version: 0.0.20
--- a/operations/experimental/windows2016-cell.yml
+++ b/operations/experimental/windows2016-cell.yml
@@ -130,30 +130,30 @@
   type: replace
   value:
     name: hwc-buildpack
-    sha1: 8f00fbd534a59137b02514127a685408e723f199
-    url: https://bosh.io/d/github.com/cloudfoundry-incubator/hwc-buildpack-release?v=2.3.11
-    version: 2.3.11
+    sha1: 7f5e6b499ae718ed858d5847190d760dda035714
+    url: https://bosh.io/d/github.com/cloudfoundry-incubator/hwc-buildpack-release?v=2.3.12
+    version: 2.3.12
 
 - path: /releases/name=winc?
   type: replace
   value:
     name: winc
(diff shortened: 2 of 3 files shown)


operations/experimental/use-offline-windows2016fs.yml
```diff
@@ -2,4 +2,4 @@
   type: replace
   value:
     name: windows2016fs
-    version: 0.0.17
+    version: 0.0.20
```
dsabeti: @sunjayBhatia, is there a way to corollate the offline windows2016fs version with windows2016fs-online-release? We're looking for a way to automate these updates?

sunjayBhatia: The online release is a regular BOSH release that we can create "final" releases from, so it can go on bosh.io but the offline release is not a regular BOSH release. It does get new git tags/github releases whenever we cut a release of it. The versions are not currently in sync because there is different code that goes into each even though they contain the same container image at a specific tag (or at least a reference to the same specific image tag).

In theory the two releases should only ever be released at the same time with an update to the container image but we have not tried to merge the versions because of the case where we need to update code in one and not another and cut a version of that release for a bug/security issue.

@dsabeti is that the info you were looking for?

dsabeti: Yeah, pretty much. I guess I'm curious then if there's a way to automatically detect if there's a new offline windowsfs version, so that we can automate the update of this ops-file. Where is the source code for the offline windowfs, and what's the process for cutting a new version?

sunjayBhatia: @dsabeti  Y'all can follow the tags/releases on this repo: https://github.com/cloudfoundry-incubator/windows2016fs-release. Whenever we cut a release we tag the latest commit with the version we want and also make a github release (https://garden-windows.ci.cf-app.com/teams/main/pipelines/windows2016fs?groups=windows2016fs-release)

The online release can of course watch bosh.io like all other releases (http://bosh.io/releases/github.com/cloudfoundry-incubator/windows2016fs-online-release?all=1)

dsabeti: Great. Thanks, Sunjay. We've got the story in flight to automatically update these releases.

sunjayBhatia: sounds great, thanks!

dsabeti:
Hey @sunjayBhatia. We've gone ahead an automated this process, so I'm going to close this PR. You should see that all of the releases are updated to latest on the `develop` branch.

PR #383, opened 2018-01-12
Update version of backup and restore sdk

edwardecook:
New version has support for MySQL 5.5, 5.6 and 5.7. Backwards compatible with the CredHub release usage.


dsabeti:
Hey @edwardecook. If you take a look at the `develop` branch, you'll see that [we've updated to the latest bbr-sdk-release](https://github.com/cloudfoundry/cf-deployment/blob/develop/operations/experimental/enable-backup-restore.yml#L6-L7). The process is now totally automated, so you won't have to submit PRs like this anymore. I'll close the PR now.

PR #394, opened 2018-02-06
Add temporary create internal domain flag 154971628

maryamklabib:
For this story: https://www.pivotaltracker.com/story/show/154971628

cc: @ameowlia


maryamklabib:
Oops this was supposed to be on develop. 

cc/ @ameowlia

PR #402, opened 2018-02-12
Update consul service_ttl to keep positive DNS results longer

ankeesler:
In some scenarios where the size of the cell is small and there are
concurrent DNS resolutions happening (e.g., running WATS in parallel),
we've noticed that the file-server.cf.internal name was not getting
resolved in time, leading to failed DNS lookups. This fixes this
scenario by making DNS cache entries live longer during the CPU spikes.

### DIFF
--- a/operations/windows-cell.yml
+++ b/operations/windows-cell.yml
@@ -17,7 +17,7 @@
         consul:
           agent:
             dns_config:
-              service_ttl: 10s
+              service_ttl: 60s
             node_name_includes_id: true
         syslog_daemon_config:
           enable: false


Deekshit51:
Hey,

is this pull request regarding the below issue ? 

https://github.com/cloudfoundry/cf-deployment/issues/403

if yes, can i retest this now ? 

Thanks in advance ,
Deekshit

ankeesler:
@Deekshit51 Hi, I do not believe that this is the same issue. This in in regards to Windows cells and your environment pertains to BOSH Lite.

ankeesler:
@ishustava Hi, we believe we were too hasty in our fix. Could we hold off in merging this change until we have completed our investigation?

ankeesler:
Thank you very much @ishustava !

PR #412, opened 2018-02-21
Removing unused variables when using an external db

garethjevans:
cc_database_password, diego_database_password & uaa_database_password are all not required when using external databases


dsabeti:
Hey @garethjevans. Can you take a look at the build failures? I think you'll need to update some fixtures for the unit tests.

garethjevans:
@dsabeti tests have been updated, those values are included in the base ops file when they probably shouldn't have been.

PR #426, opened 2018-03-05
Add ops file for enabling istio release

Gerg:
- Adds BPM because istio release requires it
- Add instance group for istio
- Enables communication between CC and istio copilot

Tracker story here: https://www.pivotaltracker.com/story/show/153187076

### DIFF
--- a/operations/experimental/README.md
+++ b/operations/experimental/README.md
@@ -17,7 +17,7 @@ and the ops-files will be removed.
 | Name | Purpose | Notes |
 |:---  |:---     |:---   |
 | [`add-credhub-lb.yml`](add-credhub-lb.yml) | Use load balancer to expose external address for CredHub. | Requires `secure-service-credentials.yml`. |
-| [`add-istio-release.yml`](add-istio-release.yml) | Add istio instance group for cloud foundry routing tier. Enables communication between capi-release and istio copilot API. | Requires bosh DNS to be enabled. |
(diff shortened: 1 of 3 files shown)


Gerg:
Note that this ops file includes BPM release because istio release depends on it. The current enable-BPM ops file enables it for a bunch of other jobs, which we don't want to make deployers do.

rosenhouse:
Hey, we're not ready to cut a final yet, so I don't think this PR should go in quite yet.

dsabeti:
Ok, I'm going to go ahead and close this github issue, but definitely submit a new one when you're ready for cf-deployment to include the experimental ops-file.

Gerg:
@rosenhouse @dsabeti Why does this ops file have to be blocked on having a final release? The release is running in routing's pipeline and routing/CAPI development environments. 

Not having a canonical ops file in cf-deployment will likely result in drift between CAPI/routing/rel int. I originally advocated for creating this ops file because there were ops files being attached to tracker stories that were different than those in the routing pipeline.

dsabeti:
@Gerg, I don't actually have a strong feeling about blocking this on having a final release available. I think it's more that the lack of the final release is a smell that there's too much active development going on to expose the ops-file. @rosenhouse and I discussed communicating this via comments / README documentation, but ultimately decided that it made more sense to wait until the release and ops-file had settled in a bit.

Having said that, in light of the risk you're describing with regards to drift between different artifacts managed by different teams, I'm happy to reconsider. It's probably best if CAPI and Routing sync up on how they want to manage this.

PR #427, opened 2018-03-05
bbl create-lbs is not a valid command

aaronjwood:
@genevieve mentioned on Slack that the correct way to do this is to follow the steps listed [here.](https://github.com/cloudfoundry/bosh-bootloader#generic-steps-for-cloud-foundry-deployment)


genevieve:
@aaronjwood Instead of removing the note entirely, maybe you can change the flags to specify it's now `--lb-type`? And also note the azure load balancers are supported.

aaronjwood:
Sure, I'll fix this up.

aaronjwood:
Sorry about the branch mix up. The other repo expected things to come from master and go back into master so I assumed the same thing for this repo.

PR #433, opened 2018-03-07
Configure ssh_proxy on BOSH-Lite to skip consul

emalm:
This change alters the disable-consul/BOSH-Lite "cross-term" operations file to configure the ssh_proxy job not to register as a Consul service. This configuration is present in the disable-consul operations file but is lost when the bosh-lite operations file adds a separate copy of the ssh_proxy job to the router and removes it from the scheduler. When Consul is actually removed from the CF deployment, the ssh-proxy then logs frequently about its failure to register.


emalm:
Thanks, @sphawley !

PR #438, opened 2018-03-13
Fix env vars in upload-stemcell command

mkb:
(With @patrick1378) Fix env vars in upload-stemcell command.


dsabeti:
Hey @mkb, thanks for the PR. I'm happy to merge the change; the only change we need is for the PR to be against the `develop` branch instead of `master`.

PR #440, opened 2018-03-15
add fast-deploy-with-downtime-and-danger.yml for impatient developers

cwlbraa:
Fresh deploys and "we're-already-experiencing-downtime" situations sometimes call for letting bosh deploy aggressively, so this ops file asks bosh to play it a bit faster and a lot looser. 

- disable canaries
- increase max_in_flight
- serial: false on everything but the ccdb
      (cloud_controller_ng seems to have stricter pre-start scripts
      than anything else in the default deployment)
- cut time for a fresh default deploy almost in half, and I haven't seen it fail yet!

using gcp and already downloaded, compiled releases:
```
bosh -n -d cf deploy cf-deployment.yml \
     -o operations/experimental/fast-deploy-with-downtime-and-danger.yml \
     -o operations/use-compiled-releases.yml
```
takes around ~17m, whereas
```
bosh -n -d cf deploy cf-deployment.yml \
     -o operations/use-compiled-releases.yml
```
takes ~31m


dsabeti:
This might be the funniest contribution to cf-deployment so far. I think it makes sense to merge, although we may steal it later for a different use case: `operations/initial-deploy.yml`.

We played around earlier with the idea of having an ops-file for only the initial deploy. The motivation at the time was to remove seed data (e.g. ASGs, quotas, UAA clients, etc.) from the manifest. But I think this ops-file would fit in line with that since downtime isn't an issue during an initial deploy. My only question about this for @cwlbraa would be this: Does the ops-file introduce and flakiness or unreliability during the initial deploy?

Either way, I'll prioritize this to get merged.

cwlbraa:
@dsabeti it doesn't add any flakiness while running it on loop with `delete-deployment` and the commands described above, but I've obviously not integrated it with everything. 

I'm cool with renaming it. it might be a bit _too_ dangerous on a loaded environment anyways, I'm not entirely sure what would happen if you tried to drain all the cells and the bbs was going down at the same time.

cwlbraa:
be forewarned, i've now seen this raise errors on aws...

dsabeti:
Womp womp.

No problem. This thing is experimental for now. I think my concern about deployment flakiness is mostly in relation to potentially co-opting this for a hypothetical feature in the future.

cwlbraa:
on AWS, scale-to-one-az + compiled releases + fast-deploy-with-danger ====> 10m CF

PR #441, opened 2018-03-19
Move BBR opsfiles to operations/backup-and-restore

flangewad:
As discussed previously with @dsabeti the BBR opsfiles are no longer considered to be experimental, so we have moved them to their own subdirectory of `operations` and updated all the READMEs.

The old file locations have been replaced with symlinks to the new locations, for backward compatibility - we'll make a new PR soon when all users have had time to migrate.

(Also minor housekeeping of other broken links we found.


flangewad:
Made an equivalent PR to develop, closing this one.

PR #442, opened 2018-03-19
 Move BRR opsfiles to operations/backup-and-restore

flangewad:
As discussed previously with @dsabeti the BBR opsfiles are no longer considered to be experimental, so we have moved them to their own subdirectory of `operations` and updated all the READMEs.

The old file locations have been replaced with symlinks to the new locations, for backward compatibility - we'll make a new PR soon when all users have had time to migrate.

(Also minor housekeeping of other broken links we found).


ChunyiLyu:
@dsabeti @staylor14 After merging this PR, could relint start automatically bumping the `backup-and-restore-sdk-release` in the update-releases pipeline? :)

dsabeti:
After discussing a bit more with the team, we actually think it would be better to keep BBR as an opt-in feature, so we're going to re-open the PR.

For context, the reasons we'd like to hold on making BBR the default are the following:
- Figuring out how/if to include DRATs in CATs.
- We've got some thoughts are changing the colocation strategy for BBR, but it would probably require some work from the release teams.

PR #443, opened 2018-03-20
Explicitly set default grootfs GC property

Callisto13:
Enables grootfs' swanky new GC behaviour by default

See: https://github.com/cloudfoundry/garden-runc-release/releases/tag/v1.12.0

[#155934143]

Signed-off-by: Tom Godkin <tgodkin@pivotal.io>

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -1080,7 +1080,6 @@ instance_groups:
         debug_listen_address: 127.0.0.1:17019
         default_container_grace_time: 0
         destroy_containers_on_start: true
-        graph_cleanup_threshold_in_mb: 0
         deny_networks:
         - 0.0.0.0/0
         persistent_image_list:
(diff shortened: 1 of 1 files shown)


ishustava:
Hey @Callisto13,

This looks good. The only thing we've noticed is that there's no need to set `rep`'s property `max_cache_size_in_bytes` to `10000000000`, as it is [the default value](https://github.com/cloudfoundry/diego-release/blob/6120e7d616db2fc71666acaa29dda05e1e20b5b4/jobs/rep/spec#L135-L137).

We're going to make this change ourselves after the merge, just wanted to give you a heads up.

Thanks!
Iryna & Ed

Callisto13:
Hey @ishustava and Ed,
We decided to explicitly use the defaults (we set the other property as the default too) just for clarity's sake. But obvs is totally your call :)

PR #453, opened 2018-03-28
Add experimental ops file to use pxc-release for a mysql database

ctaymor:
* This is for a clean install case where cf-mysql-release was never
used. Ops files for upgrading from cf-mysql-release are coming soon.

[#156242645]

Signed-off-by: Joseph Palermo <jpalermo@pivotal.io>


jpalermo:
Checks should be passing now. Anything else we need?

dsabeti:
Nope. It's prioritized to get merged soon. We'll also automate release updates and run the ops-file in our experimental environment.

ishustava:
We are not automating the release updates of the `pxc-release` for now.

For reference, here is the story to do that, as well as test the new release in our pipelines:
https://www.pivotaltracker.com/story/show/156630385

PR #464, opened 2018-04-10
Use silk-release in cf-deployment

christianang:
This PR is to use silk-release in cf-deployment instead of the silk in cf-networking-release. We have also cleaned up the experimental ops files for silk-release and updated the relevant ops files to modify silk's jobs properties instead of cf-networking jobs properties.

After this PR we will be removing silk from cf-networking-release.


dsabeti:
@christianang Don't worry about the unit test failure. It's failing because the compiled releases ops-file hasn't been updated yet. We'll do that ourselves as part of merging this PR.

christianang:
Closing this. We will be creating a new PR that combines our other PR.

Issue #429, opened 2018-03-06
operation use-gcs-blobstore-service-account.yml is not compatiable with cf-deployment.yml


alexwo:
Running Bosh interpolate with use-gcs-blobstore-service-account.yml  does not succeed.

I have bumped into the following list of issues:

1.Above operation will try to modify not existing properties in cf-deployment.yml.
There are no  "fog_connection" or  &fog_config properties defined in the cf-deployment.yml. (only blobstore-properties)

2.The dir keys properties var mappings below are also missing on the operation ->use-gcs-blobstore-service-account.yml but appear in the example file for this operation.

```
blobstore-service-account.yml` defination:
app_package_directory_key: example-apps-dir
buildpack_directory_key: example-buildpacks-dir
droplet_directory_key: example-droplets-dir
resource_directory_key: example-resources-dir
are completely missing in the operation file 
```
example:
https://github.com/cloudfoundry/cf-deployment/blob/master/operations/example-vars-files/vars-use-gcs-blobstore-service-account.yml

3.In the alternative use-gcs-blobstore.yml operation, there are "remove" sections which are missing in  -gcs-blobstore-service-account.yml but seem to be required for GCP as:

```
# remove internal blobstore including its variables
- type: remove
  path: /instance_groups/name=singleton-blobstore

- type: remove
  path: /variables/name=blobstore_admin_users_password

- type: remove
  path: /variables/name=blobstore_secure_link_secret

- type: remove
  path: /variables/name=blobstore_tls
```
https://github.com/cloudfoundry/cf-deployment/blob/master/operations/use-gcs-blobstore.yml

Your feedback is very much appreciated.

thanks,
Alex

johnsonj:
Squinting at the ops files, I wonder if it expects the `use-gcs-blobstore.yml` and `use-gcs-blobstore-service-account.yml` be applied in that order. Can @selzoc @tylerschultz confirm?

I'd expect those two to be distinct operations and if possible rename `use-gcs-blobstore-yml` to `use-gcs-blobstore-hmac.yml` (and maybe add deprecated).

selzoc:
Please refer to the [https://github.com/cloudfoundry/cf-deployment/blob/master/operations/README.md](README) which is where we document opsfile requirements.  From the entry for `use-gcs-blobstore-service-account.yml`:
```
Requires use-gcs-blobstore.yml. Introduces new variables for gcp service account credentials.
```

[...]

PR #465, opened 2018-04-11
operations: Custom S3 provider for blobstore

giner:
Can be used with non-AWS S3 providers such as minio. Based on operations/use-s3-blobstore.yml.

### DIFF
--- a/operations/README.md
+++ b/operations/README.md
@@ -23,6 +23,8 @@ This is the README for Ops-files. To learn more about `cf-deployment`, go to the
 | **Openstack** | | |
 | [`openstack.yml`](openstack.yml) | Used for deploying Cloud Foundry on OpenStack with BOSH | See [OpenStack](../iaas-support/openstack/README.md) documentation. |
 | [`use-swift-blobstore.yml`](use-swift-blobstore.yml) | Replaces local WebDAV blobstore with OpenStack swift blobstore. Used for deploying Cloud Foundry on OpenStack with BOSH | Introduces [new variables](example-vars-files/vars-use-swift-blobstore.yml) for OpenStack credentials and directory names. If you plan using the [Swift ops file](use-swift-blobstore.yml) to enable Swift as blobstore for the Cloud Controller, you should also run the [Swift extension](https://github.com/cloudfoundry-incubator/cf-openstack-validator/tree/master/extensions/object_storage). |
+| **S3-compatible** | | |
+| [`use-s3-custom-blobstore.yml`](use-s3-custom-blobstore.yml) | Replaces local WebDAV blobstore with external s3 blobstore. | Introduces [new variables](example-vars-files/vars-use-s3-custom-blobstore.yml) for s3 credentials, bucket names, etc. |
 
 
 ## Feature-based Ops-files
--- a/operations/example-vars-files/vars-use-s3-custom-blobstore.yml
+++ b/operations/example-vars-files/vars-use-s3-custom-blobstore.yml
@@ -1,0 +1,10 @@
+aws_region: us-east-1
+blobstore_access_key_id: example-access-key-id
+blobstore_secret_access_key: example-secret-access-key
+app_package_directory_key: example-app-package-directory-key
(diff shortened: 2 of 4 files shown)


heyjcollins:
Hey @giner,

Thanks for suggesting this valuable feature. 
We're happy to include it and would like to discuss a slightly different implementation which reduces ops-file redundancy and is consistent with our approach of applying ops files to cf-deployment overall.

Because the contents of `s3-custom-blobstore.yml` in your pull request is almost identical to the original `operations/use-s3-blobstore.yml` we'd like to update `s3-custom-blobstore.yml` so it contains only the replace statements for the values which differ between the two files:
```
- type: replace
  path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/cc/buildpacks/fog_connection/host
  value:
    host: ((s3_host))
- type: replace
  path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/cc/buildpacks/fog_connection/endpoint
  value:
    endpoint: ((s3_endpoint))
- type: replace
  path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/cc/buildpacks/fog_connection/path_style
  value:
    path_style: ((s3_path_style))
```
For the moment, this change would require inclusion of following operations (in order) in your `bosh deploy` command to enable a custom s3 blobstore:

`... -o operations/use-s3-blobstore.yml -o operations/s3-custom-blobstore.yml`

Any thoughts on this approach, over the one you originally provided?

Worth mentioning, there are other emerging ideas for how to ease the pain of ops-file dependencies (and other kinds of complexity). @cppforlife would be interested to see what kind of trade-offs we're discussing here. For example, potentially in the future, the call might be expressed as setting an "enable feature" option like `bosh deploy cf-deployment.yml +custom-s3`, where BOSH handles the ops-file complexity behind the scenes.

giner:
Hi @heyjcollins, I was not sure if that would be a good approach as I didn't find anything similar in the existing ops-files.

Another thing is that the key I'm changing is anchored in the original ops-file and reused through the file several times so that in s3-custom-blobstore.yml I would need to replace it not in one place but in 11(?) which may lead to mistakes in future.

dsabeti:
Good catch, @giner. 

We do actually have an example of "stacking" ops-files for external blobstores -- [`use-gcs-blobstore-service-account.yml`](https://github.com/cloudfoundry/cf-deployment/blob/master/operations/use-gcs-blobstore-service-account.yml) depends on [`use-gcs-blobstore.yml`](https://github.com/cloudfoundry/cf-deployment/blob/master/operations/use-gcs-blobstore.yml). 

**However**, [in this Github issue](https://github.com/cloudfoundry/cf-deployment/issues/429), we've gotten feedback that this model isn't very good and have a plan to try to address it. The idea is that one ops-file would swap out WebDAV for a fog connection, and require the user to provide the `fog_hash` configuration themselves. As a additional utility, we could still support ops-files for each blobstore (gcs, s3, s3-custom, azure, etc.), and those ops-files would "stack" onto the ops-file for fog connections.

I'm bringing all this up because we'd like to accept your change as is for now, but we will eventually want to make the change I've just described, so the ops-file would be significantly different at that point. The change will warrant a major version bump though, so it should be clear that things have changed.

heyjcollins:
Hello @giner  - 

I wanted to give you a heads up regarding a change we'll be making in our next point release which will impact anyone using the ops file you created in this PR.

We're getting close to publishing a cf-deployment release which alters how external blobstores are enabled. We've consolidated the portions of the ops files which were common to all the external blobstores we had ops files for previously (GCS, S3, Azure, Openstack, V-Sphere) and made it possible to enable a custom blobstore by passing in whatever values are required via a fog-connection.yml. 

These changes obviate the need for the`use-s3-custom-blobstore.yml` ops file and we'll be removing it from the set of ops files in an upcoming release of cf-deployment.

You can view the tracker story here: https://www.pivotaltracker.com/story/show/157097775

giner:
Hello @heyjcollins,

Thank you very much for sharing.

PR #467, opened 2018-04-13
Add common SANs to loggregator-agent

jasonkeene:
This allows clients of the agent to connect via:

- `localhost`
- `127.0.0.1`
- `agent`
- `metron`

This fixes an issue where the `rep` (via `go-loggregator`) is connecting to the agent and expecting a cert from the agent that has a CN or SAN of `metron`.

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -1452,6 +1452,11 @@ variables:
     extended_key_usage:
     - client_auth
     - server_auth
+    alternative_names:
+    - metron
+    - agent
+    - localhost
+    - 127.0.0.1
 - name: loggregator_tls_doppler
   type: certificate
   options:


jasonkeene:
Note: We are currently validating these changes in our environments.

dsabeti:
Hey @jasonkeene. Is this PR stale already? It looks like some commits to `develop` include similar changes.

ishustava:
Hey @jasonkeene,

Do you know why we need to add all the alternative names? It seems like cf-deployment currently works without `localhost` and `127.0.0.1`. Also, it doesn't look like `agent` (or `loggregator-agent` as we talked about in slack) is needed either.  Are you all intending to change the library to accept `agent` in the mTLS cert?

If the `go-loggregator` library is not going to change, do we still need the PR? I think we would prefer to keep SANs as strict as possible.

cc @dsabeti @staylor14

jasonkeene:
@ishustava @dsabeti Sorry, I missed your comments.

Allowing for `localhost` and `127.0.0.1` was a feature originally requested by @jvshahid . Diego was connecting to metron via `localhost`. Since they didn't have an entry in `/etc/hosts` for `127.0.0.1 metron` and no DNS record for that, they were forced to set `tls.Config.ServerName`. This works well in Go but other languages it is not so easy to do.

Allowing for `agent` or `loggregator-agent` was part of moving away from the name `metron`. We have moved away from that name internally and want to be open to transitioning users away from it as well.

dsabeti:
Thanks for the update. Is there any action we need to take on our side? Since we decided to include the metron_agent rename in a major version bump, would we need to re-introduce these changes at that point in time?

jasonkeene:
Also, @dsabeti I just checked develop and master and didn't see any SANs for the loggregator agent.

PR #469, opened 2018-04-17
Ops file for external s3 unversioned blobstore backup and restore

MirahImage:
There is an [open PR](https://github.com/cloudfoundry/cf-deployment/pull/442) to move the existing backup and restore operations files out of the `operations/experimental` directory. We think they will land in the `operations/` directory, so we have put this new operations file there.

Mirah and @jamesjoshuahill


gcapizzi:
Closing this PR and including this new ops file in #442

PR #471, opened 2018-04-18
add denamespaced properties for cf-networking-release jobs

tylerschultz:
This duplicated the properties so that we can move to denamespaced properties. This change will allow us to cut a new release without disrupting the pipeline for cf-deployment. After we release we will make a new PR to remove the duplicate (namespaced) properties and everything should be happy.

next steps:
  1. cut a final release that uses these properties
  2. remove the old properties from cf-deployment

[#156443433]

Signed-off-by: Gabe Rosenhouse <grosenhouse@pivotal.io>


tylerschultz:
closing, we're regrouping. we'll be back with another.

PR #475, opened 2018-04-25
Rename bits-service URLs from bits-service.* to bits.*

petergtz:
`bits-service` as a name for URLs doesn't make a lot of sense, as we're dealing with `bits` and that's what the URL should describe. Also, `bits-service.service.cf.internal` always looked repetitive and silly. Therefore changing it.

This change requires an update in certificates. We decided now is still a better time to make such an invasive change than when it is out of experimental and potentially deployed in many production environments.

### DIFF
--- a/operations/experimental/bits-service.yml
+++ b/operations/experimental/bits-service.yml
@@ -12,8 +12,8 @@
           client_key: ((cc_tls.private_key))
         droplets: null
         packages: null
-        private_endpoint: https://bits-service.service.cf.internal
-        public_endpoint: https://bits-service.((system_domain))
+        private_endpoint: https://bits.service.cf.internal
+        public_endpoint: https://bits.((system_domain))
         secret: ((bits_service_secret))
         signing_users:
         - password: ((bits_service_signing_password))
@@ -24,20 +24,20 @@
   value:
     name: bits-service
     registration_interval: 20s
-    server_cert_domain_san: https://bits-service.((system_domain))
+    server_cert_domain_san: https://bits.((system_domain))
     tags:
       component: bits-service
     tls_port: 443
     uris:
-    - bits-service.((system_domain))
+    - bits.((system_domain))
 - type: replace
   path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/cc/bits_service?
   value:
     ca_cert: ((service_cf_internal_ca.certificate))
     enabled: true
     password: ((bits_service_signing_password))
-    private_endpoint: https://bits-service.service.cf.internal
-    public_endpoint: https://bits-service.((system_domain))
+    private_endpoint: https://bits.service.cf.internal
+    public_endpoint: https://bits.((system_domain))
     username: admin
 - type: replace
   path: /instance_groups/name=cc-worker/jobs/name=cloud_controller_worker/properties/cc/bits_service?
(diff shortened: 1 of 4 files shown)


ishustava:
Hey @petergtz,

We have some questions about this change:
1. Since BOSH vars store and CredHub don't regenerate variables if parameters change, it means that `bits_service_ssl` variable will still have old common name and alternative names, unless it's manually deleted prior to deploy. The only way around it, without manual intervention, would be to rename the variable.
2. Do you know if changing bosh dns/consul URLs would cause downtime during the deploy? Maybe, for the BOSH DNS ops file we could keep both aliases to avoid this potential downtime, but I'm not quite sure how to change a consul URL without downtime.

petergtz:
Hi @ishustava,

> Do you know if changing bosh dns/consul URLs would cause downtime during the deploy?

Yes, it probably will. Your deployment might not even go through, because CC will wait until it can access the Bits-Service, which in this case will fail.

I see 2 options to roll out this change:
1. Since you guys deactivated the Bits-Service temporarily anyway, because it caused some issues with CATs for you, you could simply remove the `bits_service_ssl` variable completely, while the Bits-Service is deactivated. Once you re-activate it again (when we have resolved the CATs issues), your deployment will re-generate this variable appropriately.
2. I could change the PR to allow both names (`bits.` and `bits-service.`) in the certificate. You'd still have to manually delete `bits_service_ssl` to get it regenerated with the next deployment, but it seems that would smoothen the roll-out.

I'd vote for option 1, because it is less work for both sides and seems to be good timing with the current conditions.

dsabeti:
@petergtz, I'm thinking about anybody who has bits-service deployed already. If we go with option 1, is there any work-around for them? I think it's OK if there's a little bit of downtime, given that the feature is still experimental, but I just want to make sure nobody ends up stuck.

petergtz:
@dsabeti 

> I'm thinking about anybody who has bits-service deployed already. 

Good point. I implicitly assumed that you guys and CloudOps are the only ones using Bits-Service. But I should not make that assumption.

> I think it's OK if there's a little bit of downtime, given that the feature is still experimental, but I just want to make sure nobody ends up stuck.

Also a good point. I think that deployment could get stuck. I didn't bother in our environments to see what the problem was, simply nuked them. Given that we update the endpoint and certificates will therefore be invalid, I think CC will not be able to connect to Bits-Service (as mentioned in https://github.com/cloudfoundry/cf-deployment/pull/475#issuecomment-385925119).

To roll this change out, we should do the following steps/separate deployments:
1. Recreate the `bits_service_ssl` certificate and allow `bits.*` and `bits-service.*`.
2. Change all URLs to use `bits.*`
3. Optionally recreate the `bits_service_ssl` certificate to only allow `bits.*`.

I can provide the necessary PRs, but how would we provide this to users?

dsabeti:
@petergtz I think if you submit the PR to allow for both `bits.*` and `bits-service.*`, that should probably be enough. In the release notes that include that change, we'll advertise that operators using the bits-service should regenerate their certificates. We can warn that, when the bits-service is GA'd some time in the future, we'll use only `bits.*`, so they'll need to regenerate by then.

petergtz:
@dsabeti Makes sense. Latest commit makes the certificate valid for both as you suggested.

dsabeti:
Hey @petergtz. It looks like most the components still use `bits.*` instead of `bits-service.*`. In your list above, it achieves steps 1 and 2 at the same time. However, I think this changes need to only update the certificates. In a subsequent release, we'd switch from `bits-service.*` to `bits.*`.

petergtz:
@dsabeti Oh, OK, sorry. I misunderstood you. Now it's clear. Will make the change.

petergtz:
@dsabeti I've created a separate PR for this, to keep the commit history simple: https://github.com/cloudfoundry/cf-deployment/pull/482

dsabeti:
Thanks @petergtz. Sorry for all the back-and-forth.

PR #481, opened 2018-05-02
Add new required encryption_keys property to uaa job

DennisDenuto:
[#157011310] https://www.pivotaltracker.com/story/show/157011310

Signed-off-by: Jaskanwal Pawar <jaskanwal.pawar@gmail.com>


tnwang:
As background - this is for the UAA v58 release which will require an encryption key to be set.

Issue #411, opened 2018-02-20
App log is not displaying with cf logs 


lordcf:
Hi,
We have deployed the CF using cf-deployment repo. All the VMs are in running state but the application logs is not displaying.

tried to get logs using:
cf logs APP_NAME
cf logs APP_NAME --recent

cf-deployment manifest version: v1.12.0
cloud: Alibaba

I have attached the screen shot.
![image](https://user-images.githubusercontent.com/25821409/36430399-1643fd14-167b-11e8-96d5-e11e37f4f41a.png)

dsabeti:
Hi @lordcf. While you're running `cf logs`, can you also make an HTTP request (perhaps with `curl`) against the app? That should force some logs through.

I'm also going to loop in @xiaozhu36, who helps to support cf-deployment on Alibaba Cloud.

lordcf:
Hi @dsabeti . I have curl the application after that too cf logs is not displaying anything.
The message will be like this only.

"Retrieving logs for app php-demo in org system / space testing as testuser...

dsabeti:
@lordcf, ok. It looks like the traffic controller can't connect to the dopplers.  Can you check that there aren't any firewall rules blocking networking traffic between the two IP addresses?

Let's also get help from the Loggregator team: cc @ahevenor @JohannaSmith

JohannaSmith:
@lordcf Those are pretty standard logs as the trafficcontroller continually reconnects to dopplers. Based on the time gap between the connection log and the gRPC context cancellation, I do believe trafficcontroller is discovering the doppler. 
I notice that it only connects to 1 doppler. How many doppler instances and trafficcontroller instances do you have? Because we would recommend at least 2 dopplers per AZ to increase log reliability/ chance of it getting through the system.

lordcf:
@JohannaSmith  : Thanks for response.
We found the issue. Actually there is an metron_agent addons configure in the https://github.com/cloudfoundry/cf-deployment/blob/master/cf-deployment.yml#L10 files. But while deploying the cf. its not applying the by default to all vms (no metron_agent job running on the VMs).  We have created the new file with metron_agent.yml  which configuration is mention on cf-deployment.  After that I have done "update-runtime-config"  and then redeployed CF. 
Now metron_agent job is running on each VM and we are getting the logs.

But we have query, Why menton_agent addons is not applying while first time cf deployment?

dsabeti:
[...]

PR #485, opened 2018-05-05
fix no app log bug

xiaozhu36:
This PR aims to fix [issue 441](https://github.com/cloudfoundry/cf-deployment/issues/411) that there is no output logs while pushing a application.

### DIFF
--- a/iaas-support/alicloud/stemcells.yml
+++ b/iaas-support/alicloud/stemcells.yml
@@ -1,12 +1,5 @@
 ---
 # Light Stemcells
-- type: remove
-  path: /stemcells/alias=default/os?
-
-- type: replace
-  path: /stemcells/alias=default/name?
-  value: bosh-alicloud-kvm-ubuntu-trusty-go_agent
-
 - type: replace
   path: /stemcells/alias=default/version?
   value: 1016


dsabeti:
Hi @xiaozhu36. We're happy to merge this, but I'm curious how this addresses the problem in the Github issue you referenced.

xiaozhu36:
HI @dsabeti , Thanks for your reply. Actually, I have not found the actual reason which resulting in the problem. This PR comes from our customer's one times test.

heyjcollins:
Hi @xiaozhu36 - have you confirmed the change you've submitted in this PR does resolve the customer problem? Can you provide more detail? You also mentioned this was from a "customer's one times test."
Is this change still needed?

xiaozhu36:
HI @heyjcollins I am sorry and I think I gave you a wrong message. This PR is not related to the [issue "App log is not displaying with cf logs"](https://github.com/cloudfoundry/cf-deployment/issues/411) and I have found the really reason that I need to modify ntp to China timezone in alibaba cloud [CPI manifest](https://github.com/cloudfoundry/bosh-deployment/pull/175/files#diff-b53e17bd6d341a400de58b1e62e4f693R110). I will close the PR. Thanks a lot.

heyjcollins:
Hi @xiaozhu36,
Thanks for responding. I'm glad you were able to resolve the issue.

PR #487, opened 2018-05-07
add params for port & signature version

jyriok:
as i use a s3-like with ceph, i need a little more params like  port and aws_signature_version (my actual storage is only compatible with v2 signature) 
thanks !


dsabeti:
Hi @jyriok. We're going to completely change the way we provide support for external blobstores soon. You can take a look at the stories leading up to [this release marker](https://www.pivotaltracker.com/story/show/155745219) for more information, but the shot description is that we'll provide an ops-file for operators to bring their own customized `fog_connection`. As such, I'm inclined to close the PR, but I'm happy to discuss more.

jyriok:
no pb for me ! for the moment,  i've custom ops file  but i'll be happy to use your new ops file when it'll be ready :)

PR #488, opened 2018-05-08
Remove redundant UAA port from Diego SSH-proxy job

emalm:
Starting with [Diego v2.5.0](https://github.com/cloudfoundry/diego-release/releases/tag/v2.5.0), the default value for this property is 8443, so this manifest property is now redundant.

[#155735780]


emalm:
Thanks, @ishustava !

PR #491, opened 2018-05-08
Cf networking 2.0

adobley:
add opsfiles for cf-networking/silk 2.x

Use pinned versions for networking and silk


ishustava:
Hey @adobley & @christianang,

Looks like some unit tests are failing. Do you mind taking a look?

ishustava:
Also-- it looks like this PR makes changes to other ops-files. Could you move those changes to new ops-files that are also in the `experimental` directory? Otherwise, operators who are using the `use-postgres.yml`, etc., ops-files but not yours will get a bunch of properties added to their manifest that they aren't using.

As an example, we're currently doing a similar thing with `secure-service-credentials-with-pxc-release.yml` or `secure-service-credentials-postgres.yml`. In your case, you could create a new experimental ops file named `use-external-dbs-with-cf-networking-2.yml` or something similar.

adobley:
Alrighty. We'll make the duplicate opsfiles for those. Thanks!

dsabeti:
@ishustava, I think the new properties should actually be fine -- each release will ignore the properties they don't use.

ishustava:
Hey @adobley,

just to clarify, I meant to say to add necessary properties to the future `use-external-dbs-with-cf-networking-2.yml` and not duplicate them.

dsabeti:
@ishustava @adobley, let's chat on Slack tomorrow morning about this

ishustava:
@dsabeti 

True, but the ops files won't interpolate without the `?`.  If you add `?`, you will end up with extra properties even if you're not using cf-networking 2.0.

Happy to chat tomorrow though :)

ishustava:
Merged into develop with this commit: `f3db8c0b824f24d4f811a9827b03354bbc985b79`.

PR #495, opened 2018-05-15
Add perm.admin scope to the CF CLI admin user

oozie:
Hi,

We want to add `perm.admin` scope in 2 places in the `cf-deployment.yml`, one under `uaa/scim/users/name=admin` and the other to the comma-separated list of scopes under CF CLI admin user.

We can accomplish the former with an opsfile. If we use an opsfile for the latter, we'll have to keep it in sync with `cf-deployment.yml`'s value. 

One way around that is to append the scope in the mainline `cf-deployment.yml` file which this PR illustrates.

Are you okay with us doing it that way, or should we hardcode the long list of existing scopes into the opsfile?

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -413,7 +413,7 @@ instance_groups:
             authorized-grant-types: password,refresh_token
             override: true
             refresh-token-validity: 2592000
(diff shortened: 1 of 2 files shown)


bgandon:
I submitted in cloudfoundry/uaa-release/pull/89 a patch on the `uaa-realease` for adding support for a YAML list in a new `scopes` property of UAA clients definitions.

oozie:
Wow, thanks @bgandon for your help. We're going to close this PR because we targeted it at the wrong branch, and we're going to reopen a similar one against develop.
For now, we'll resubmit the same request not to block the merge of this on the UAA release getting into `cf-deployment`.

PR #496, opened 2018-05-15
pxc updates

jpalermo:
- Update pxc version in migration ops file.

- Change pxc certificates to use service_cf_internal_ca CA rather than creating their own.

Any existing deployments will be fine since the certificates won't get recreated. If there is a desire to update an existing deployment, it will need to delete the existing certificate so it is regenerated as well as scale down to a single database instance.

### DIFF
--- a/operations/experimental/migrate-cf-mysql-to-pxc.yml
+++ b/operations/experimental/migrate-cf-mysql-to-pxc.yml
@@ -126,28 +126,10 @@
 - type: replace
   path: /variables/-
   value:
-    name: pxc_galera_ca
-    type: certificate
-    options:
-      is_ca: true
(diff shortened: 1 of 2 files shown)


heyjcollins:
Hey @jpalermo  - We considered the UX involved with using the shared vs pxc-specific CA and decided to go with pxc-specific.
I'm going to close this PR. Please submit another PR which for incrementing your pxc release from 0.4.0 to 0.6.0 and we'll integrate that. Cheers!

emalm:
@heyjcollins I'm curious about the UX considerations that led us to prefer the separate pxc-specific CA instead of the shared internal CA: in isolation, I would think that introducing yet another CA in cf-deployment would be more confusing...

PR #502, opened 2018-05-18
Add db tls for pxc

ctaymor:
Enable Core CF components to talk to PXC using TLS
- Add addresses for cc, diego, routing, and uaa so that they use
hostnames which match the mysql_server_cert's common name
- Pass the pxc_server_ca to cc, diego, and routing via bosh properties
- Add the ca_certs job from os-conf release to the uaa instance group, and use that job to pass the pxc_server_ca to uaa
- Turn on DB TLS for components which require it to be explicitly
enabled

Note that this does not turn on TLS for cf-networking. The ability to pass cf-networking the DB CA is only enabled in cf-networking 2.0, which is not yet in cf-deployment.

UAA does not support passing the CA via a bosh property, and recommended using the os-conf release to pass the CA.

### DIFF
--- a/operations/experimental/migrate-cf-mysql-to-pxc.yml
+++ b/operations/experimental/migrate-cf-mysql-to-pxc.yml
@@ -8,7 +8,7 @@
     version: 0.6.0
 
 - type: replace
-  path: /releases/-
+  path: /releases/name=os-conf?
   value:
     name: os-conf
     version: 20
--- a/operations/experimental/migrate-cf-mysql-to-pxc.yml
+++ b/operations/experimental/migrate-cf-mysql-to-pxc.yml
@@ -8,6 +8,14 @@
     version: 0.6.0
 
 - type: replace
+  path: /releases/-
+  value:
+    name: os-conf
+    version: 20
+    url: https://bosh.io/d/github.com/cloudfoundry/os-conf-release?v=20
+    sha1: 42b1295896c1fbcd36b55bfdedfe86782b2c9fba
+
+- type: replace
   path: /instance_groups/name=database/jobs/name=mysql/properties/cf_mysql_enabled?
   value: false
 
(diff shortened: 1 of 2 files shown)


operations/experimental/migrate-cf-mysql-to-pxc.yml
```diff
@@ -3,9 +3,17 @@
...
     name: pxc
-    sha1: bd784bb93ffeff62cdec6aeb62ba0fea4eba6f41
-    url: https://bosh.io/d/github.com/cloudfoundry-incubator/pxc-release?v=0.4.0
-    version: 0.4.0
+    sha1: 4dd970acd8fb059e73d9922adb9cc40206334882
+    url: https://bosh.io/d/github.com/cloudfoundry-incubator/pxc-release?v=0.6.0
+    version: 0.6.0
+
+- type: replace
+  path: /releases/-
```
dsabeti: Can we change this to a update-or-append operator? I'm worried about the case where an operator has added os-conf-release in a runtime-config or something. I thing it would need to be something like
```
- type: replace
  path: /releases/name=os-conf?
  value:
  ...
```

operations/experimental/migrate-cf-mysql-to-pxc.yml
```diff
@@ -159,3 +167,63 @@
...
+  value: "((pxc_server_ca.certificate))"
+
+- type: replace
+  path: /instance_groups/name=diego-api/jobs/name=locket/properties/diego/locket/sql/require_ssl?
+  value: true
+
+- type: replace
+  path: /instance_groups/name=uaa/jobs/-
+  value:
+    name: ca_certs
```
dsabeti: Similarly, I wonder if this should also have a guard in place:
```
- type: replace
  path: /instance_groups/name=uaa/jobs/name=ca_certs?
  value:
  ...
```

Or even:
```
- type: replace
  path: /instance_groups/name=uaa/jobs/name=ca_certs?
  value:
  ...
```

operations/experimental/use-pxc.yml
```diff
@@ -5,12 +5,23 @@
     sha1: 4dd970acd8fb059e73d9922adb9cc40206334882
     url: https://bosh.io/d/github.com/cloudfoundry-incubator/pxc-release?v=0.6.0
     version: 0.6.0
+
+- type: replace
+  path: /releases/-
+  value:
+    name: os-conf
```
dsabeti: See above.

operations/experimental/use-pxc.yml
```diff
@@ -120,3 +140,63 @@
...
+  value: "((pxc_server_ca.certificate))"
+
+- type: replace
+  path: /instance_groups/name=diego-api/jobs/name=locket/properties/diego/locket/sql/require_ssl?
+  value: true
+
+- type: replace
+  path: /instance_groups/name=uaa/jobs/-
+  value:
+    name: ca_certs
```
dsabeti: See above

operations/experimental/migrate-cf-mysql-to-pxc.yml
ctaymor: @jpalermo and I think the current implementation (below) is actually safer. The os-conf job does not take a list of certs, it requires them to all be concatenated into a single string. If the operator gets two ca_certs jobs, the deploy should fail fast, and hopefully they can figure out how to add an ops file with all the certs they need to provide to UAA concatenated.
```
- type: replace
  path: /instance_groups/name=uaa/jobs/-
  value:
    name: ca_certs
    release: os-conf
    properties:
      certs: "((pxc_server_ca.certificate))"
```

If we instead add the guard (`path: /instance_groups/name=uaa/jobs/name=ca_certs?`), then if the operator has two copies of this job, it won't fail until after deployment when something (either the DB or the job using the other cert) fails to talk to UAA. That sounds like a more painful debugging experience.

We opened an issue for the os-conf release about fixing this in the future, but it seems like the mediocre but best option for now. https://github.com/cloudfoundry/os-conf-release/issues/34

operations/experimental/migrate-cf-mysql-to-pxc.yml
ctaymor: Sounds good. Pushed a commit to update that.

operations/experimental/migrate-cf-mysql-to-pxc.yml
dsabeti: Heh, I realize now that my second implementation suggestion is identical to the first. I think I meant this:
```
- type: replace
  path: /instance_groups/name=uaa/jobs/name=ca_certs?
  value:
    name: ca_certs
    release: os-conf

- type: replace
  path: /instance_groups/name=uaa/jobs/name=ca_certs/properties
  value:
    certs: ...
```

However, in light of your comment that the certs can't be appended, it seems reasonable to leave it as you originally implemented it.

PR #505, opened 2018-05-23
Update pxc database ports to match cf-mysql port

ndhanushkodi:
We noticed that the `use-pxc.yml` file was using a different port than `cf-deployment.yml` and the `migrate-cf-mysql-to-pxc.yml` ops file.


ishustava:
@ndhanushkodi 

This looks great! 

Would you mind publicizing your GitHub Org membership as gitbot suggests above? You might need to re-open the PR again to make gitbot happy.

PR #507, opened 2018-05-25
Make database in blobstore optional in rename-network

Ykisialiou:
Database and blobstore can be changed to external by other ops-files (and it's one of our cases). So I think it's better to have corresponding instance_groups optional in `rename-network.yml` ops-file to avoid  
`Expected to find exactly one matching array item for path` errors and too strict ops-files order.

### DIFF
--- a/operations/rename-network.yml
+++ b/operations/rename-network.yml
@@ -19,7 +19,7 @@
   value: ((network_name))
 
 - type: replace
-  path: /instance_groups/name=database/networks/name=default/name
+  path: /instance_groups/name=database?/networks/name=default/name
   value: ((network_name))
 
 - type: replace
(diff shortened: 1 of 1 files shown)


heyjcollins:
Hey @Ykisialiou - We attempted to support operator intention to rename their network and admit this implementation is fragile as evidenced by the use case you bring up above. 

The solution you recommend will result in the undesired _creation_ of a a database entry in your cf-deployment manifest rather than simply ignoring it when it doesn't exist because it's external.

Unfortunately we don't have a graceful ops-file-based solution to this problem and will bring this feedback to the bosh team to consider.

heyjcollins:
Hey @Ykisialiou,

I'm doing some PR grooming and will be closing this PR as it's not something we can implement.
If you'd like to continue the dialogue, please feel free to reopen it.

Thanks so much,

Josh

PR #511, opened 2018-05-30
Remove UAA tls for pxc

APShirley:
- UAA often fails to start if `os-conf` does not add the cert fast enough (race condition)

[#157974601]


ishustava:
Hey @APShirley,

Thanks for this PR!  I have a few questions/suggestions:

- Looks like with this change we no longer need to add `os-conf-release`. Would you mind updating the PR to remove it?
- `migrate-cf-mysql-to-pxc.yml` still enables TLS on UAA with `os-conf`. Is this desired behavior? If not, that ops file would also need to remove `os-conf-release`, in addition to disabling TLS for UAA.

Thanks!
Iryna

ctaymor:
Hey @ishustava, yeah both of those should be removed. We updated the PR with them.

PR #512, opened 2018-05-30
Move nfs broker to cf-pushed model to reduce VM count

julian-hj:
#### commits:
- update to the latest version of nfs-volume-release and switch service broker to run as a cf pushed app instead of a bosh job running on its own vm. [#155384701]


heyjcollins:
Hey @julian-hj  - Looks like some unit test(s) failed due to your changes. We've pulled this into our backlog for this week and are hoping you can resolve the unit test issues as soon as possible so we can proceed. THANKS

julian-hj:
@heyjcollins I did look at that, but I am at a loss as to why this PR would cause the failure in question.  We don't use silk-release.  Is there a way you could kick the units again?  Or if you have more light to shed on the failure that would be helpful too...

julian-hj:
I do see that the ops file lacks a trailing line feed.  I've fixed that just now.  Maybe that will solve the issue...

julian-hj:
@heyjcollins -- no difference with the trailing whitespace fixed.  The units are complaining that they cannot download https://bosh.io/d/github.com/cloudfoundry/silk-release?v=2.2.0 which downloads from a browser just fine.  Thoughts?

heyjcollins:
@julian-hj - I'll have someone on the team here take a looksy.

ishustava:
The changes in this PR have been merged in this commit: `faede4da`

PR #516, opened 2018-06-04
Move bbr-credhubdb job to bbr instance group

mdelillo:
We moved the `bbr-credhub` job to the `backup-restore` instance group with the rest of the BBR jbos.

We also removed the unnecessary opsfiles. Credhub 1.9.2 supports using bosh links for `data_storage`, so we no longer need an opsfile per database we support.


crawsible:
Hey @dsabeti @ishustava, any questions for us on this PR?

ishustava:
Hey @crawsible @mdelillo,

This looks good, thanks! Pinging @heyjcollins to prioritize.

My only comment would be to keep the deleted files empty rather than deleting them for backwards compatibility. If you get a chance to update this PR before we get to it, that would be great. Otherwise, we can make the change when we merge.

PR #517, opened 2018-06-04
Add the UAA section and hints to notes

oozie:
`pxc_server_ca` reference works only with the PXC release. `mysql_server_certificate.ca` is available in both. This change makes the opsfile usable with both PXC and regular mysql release.

[#157527204]


oozie:
Thanks @ishustava for reaching out to fix the PR. I renamed the title to better reflect it current state. The PR now only includes a new UAA section and deployment hints reference.

oozie:
Greetings! How do yous feel about merging this in? We have a few stories to deliver that are blocked on this.

PR #519, opened 2018-06-05
Make the client id match its definition (dash/underscore)

oozie:
This bug prevents migrator from working because dashed-client-id does not match underscored_client_id in the other section of the manifest.

[#157952485]

### DIFF
--- a/operations/experimental/perm-service.yml
+++ b/operations/experimental/perm-service.yml
@@ -138,7 +138,7 @@
           hostname: uaa.((system_domain))
           port: 443
           ca_certs:
-          - ((router_ssl.ca))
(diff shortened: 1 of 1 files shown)


staylor14:
Hey @oozie!  In cf-deployment.yml, we typically use underscores for the names of Bosh variables, and dashes for instance groups.  I don't know what the `migrator` is, but can it be made to follow this (unwritten and unexplained) standard?

Thanks!  @heyjcollins && Steve

oozie:
Ack, @staylor14 I forced pushed a removal of the last commit. I hope the PR looks good now.

heyjcollins:
@oozie - looks good now yes.

PR #525, opened 2018-06-10
Drop unused locket database link.

jmcarp:
Under the latest version of bosh, setting a link that isn't specified in the job spec raises an error, like this: `Job 'locket' in instance group 'diego-api' specifies consumers in the manifest but the job does not define any consumers in the release spec`. This patch drops the unused database link from the locket job in the external database operations file.

### DIFF
--- a/operations/use-external-dbs.yml
+++ b/operations/use-external-dbs.yml
@@ -220,6 +220,3 @@
 - type: replace
   path: /instance_groups/name=api/jobs/name=policy-server/consumes?
   value: {database: nil}
-- type: replace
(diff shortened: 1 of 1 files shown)


aeijdenberg:
And I was just about to open the same PR!

"Me too" from us, this has broken our deployments, so for now we're also removing via another operator file.

Thanks @jmcarp - I'm going to drop a note in the cf-deployment Slack channel as well as this might bite other others (since as of a few hours ago, the latest bosh has hit `bosh-deployment`)

jmcarp:
cc @dsabeti. It would be nice to get this included in the next release.

PR #529, opened 2018-06-14
Deprecate and void obsolete secure-rep ops files

emalm:
Diego v2.0.0 removes the `admin_api.require_tls` property from the `rep`
and `rep_windows` jobs in favor of always enabling TLS on the rep admin
API server. Consequently, these operations files that use the property
to secure the rep API have no effect.

Note that this is suitable for incorporation into cf-deployment v1.x as
it maintains backwards compatibility with the cf-deployment filesystem
interface. A separate PR will follow to remove these files entirely
for cf-deployment v2.0.


heyjcollins:
Thanks @emalm  --- we won't have another 1.x release of cf-deployment. 
We're releasing cf-d 2.0.0 on Monday. 
How would you like to proceed?

emalm:
@heyjcollins No problem, we can pull this deprecation change into the v2.x line and then remove the ops files entirely in v3.0.0. From watching the Rel Int team prepare for v2.0.0, I expect the deprecation notice and empty ops files should suffice to ensure their removal on the next major version, and so I'll dispense with an explicit PR to remove them.

Thanks,
Eric

heyjcollins:
@emalm - ok good. We'll pull this PR into v2.1.0 and carry on as you described.
Cheers!

heyjcollins:
@emalm - thanks for bumpage. Yes we'll publish before 3.0

PR #531, opened 2018-06-14
Support to stemcell version 1018

xiaozhu36:
This PR change stemcell version to 1018 and it can support bosh blobstore with alibaba cloud oss.


xiaozhu36:
Hi @heyjcollins At present, our code [PR ](https://github.com/cloudfoundry/bosh-linux-stemcell-builder/pull/24) has not been merged by official, so v3586.18 does not support alibaba cloud. If you want to deploy cf on alibaba cloud, you can use our self-building stemcell:

PR #533, opened 2018-06-14
Add cf networking components to pxc opsfiles

christianang:
This PR updates the PXC opsfiles to update policy-server and silk-controller to use TLS connections with PXC.

/cc @tylerschultz


tylerschultz:
cc @APShirley

PR #535, opened 2018-06-15
Log cache promote

MasslessParticle:
This is a PR to move Log Cache into cf-deployment proper


heyjcollins:
Hi @MasslessParticle - I see the PR failed 1 of our checks. We'd like to integrate these changes and would like to make sure this failure is resolved. 

Additionally, we're enabling log-cache in our CATs for our experimental environment and would like to have that run for at least a week before we make log-cache default in cf-d

heyjcollins:
Hey @MasslessParticle - I chatted with @ahevenor and mentioned I'd be happy to pull this in once it's proven to work with and without BOSH DNS enabled - and heard back that Log-Cache requires bosh-dns so we'll need to block on that unless y'all fix the issue.

Additionally, since the default implementation here requires a new instance group, I'd like to also provide a couple options to operators who may not have heavily loaded environments or who would rather not use log-cache at all.

RelInt is working to default bosh-dns in the next few weeks so could the loggregator team update this PR with a couple additional ops files:

`disable-log-cache.yml`
`colocate-log-cache-with-doppler.yml` --- or whatever colocation strategy you believe will be optimal

Thank you!

heyjcollins:
update on previous request - since the intention now is to roll out with log-cache colocated with doppler, there's no need to include the `colocate-log-cache-with-doppler.yml` ops file with this PR.

cc @staylor14

toddboom:
@heyjcollins Ok, we pushed up the ops file to disable log cache and added it to the interpolation test script, plus we bumped to the v1.4.0 release. It looks like we're still failing on compiled releases and a bosh-dns error that doesn't look like our fault, but please let us know if you need any additional changes/cleanup to get this merged in!

/cc @hdub2

PR #537, opened 2018-06-18
Add UAA TLS

ctaymor:
UAA v60 now supports accepting certs for TLS to the DB via a bosh property. This PR enables TLS from UAA to the mysql DB.

It also changes the mysql CA Cert to be referenced through the `mysql_server_certificate` variable's CA property. This is the same CA as the previous `pxc_server_ca` certificate property, but referenced in a way that enables easier CA rotation.

### DIFF
--- a/operations/experimental/migrate-cf-mysql-to-pxc.yml
+++ b/operations/experimental/migrate-cf-mysql-to-pxc.yml
@@ -155,6 +155,9 @@
   path: /instance_groups/name=scheduler/jobs/name=cloud_controller_clock/properties/ccdb/address?
   value: sql-db.service.cf.internal
 - type: replace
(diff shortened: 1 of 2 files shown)


ishustava:
Hey @ctaymor,

Thanks for this PR! 

A while ago we changed all CA properties to be referenced from the CA variable, and it was also done to enable easier rotation(#305) haha!

Is there a reason you believe the CA rotation will be easier if CAs are references from their certificate variables?

APShirley:
Let's update this PR and remove the ca variable references. We can open another PR if we want to keep our way of referencing variables. cc @jpalermo

APShirley:
We've updated the PR to not change how certs are referenced

ishustava:
Looks good! thanks @APShirley!

PR #539, opened 2018-06-19
Move service discovery and bosh dns opsfiles

christianang:
- no longer in experimental (wooooooooo)

[#158277374]

Signed-off-by: Amelia Downs <adowns@pivotal.io>


heyjcollins:
Hi Networking Friends,

After discussing enabling bosh DNS and service discovery further with Dieu and the RelInt team, I've changed my mind about this approach and I'd like to go directly to in-lining rather than promoting from operations/experimental to operations.

Could you update this PR so it 
1. inlines to cf-deployment rather than promotes out of experimental?
1. keep the experimental ops files in-place but empty to support any operators currently using those files in their deployments 
1. update  `operations/experimental/README.md` for the ops files to state they are deprecated since the feature's been GA'd and the files will be removed in cf-d 3.0.

Apologies for the change in direction.

PR #541, opened 2018-06-21
Place the ca_certs property in the tls namespace

oozie:
perm-release spec addresses UAA's ca certs by `uaa.tls.ca_certs`.

This fix will make the deployment with `auth` feature-flag succeed.

[#158535821]


oozie:
Friendly ping! I think the change is safe for merge.

PR #551, opened 2018-07-06
mysql proxy job uses regular consul service discovery

jpalermo:

We are removing outdated consul code from the mysql proxy job. This PR changes cf-deployment to use regular consul service discovery for consul instead of the special code inside the proxy.

- [X] YES 
- [ ] NO

Update database proxy job to use consul service discovery in a way more consistent with other jobs.

No

- [X] NO

- [X] GA'd feature/component


jpalermo:
Realized we are missing some pieces, we'll make this PR when we cut the release and have the dns_health_check

PR #564, opened 2018-07-17
Increase the route registrar's cc timeout on bosh-lite

ericpromislow:

We have observed in some CI systems that the route registrar will
deregister the single instance of a cloud controller if that cloud
controller is under load and cannot respond to a healthcheck in time.

So, this commit will increase the amount of time that a registrar will
allow for the cloud controller to respond.

[#158990935](https://www.pivotaltracker.com/story/show/158990935)

- [x] YES 
- [] NO

Updated `operations/bosh-lite.yml`

- Lengthened the route registrar's default cloud controller healthcheck timeout from 3 seconds to 10 seconds. We have observed that some bosh-lites under load will deregister their cloud controller.

No.

- [x] NO

- [ ] GA'd feature/component

- [ x] **Slightly Less than Urgent**

/cc @mkenyon @selzoc @tcdowney @zrob

### DIFF
--- a/operations/bosh-lite.yml
+++ b/operations/bosh-lite.yml
@@ -102,6 +102,11 @@
   path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/cc/default_app_memory?
   value: 256
 
+# ----- Increase route registrar's CC timeout, in case of load ------
+- type: replace
+  path: /instance_groups/name=api/jobs/name=route_registrar/properties/route_registrar/routes/name=api/health_check/timeout?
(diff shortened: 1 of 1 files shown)


tcdowney:
We closed this because this behavior affects more than just bosh-lite. We've created a new PR to tweak the value in the main `cf-deployment.yml` manifest: https://github.com/cloudfoundry/cf-deployment/pull/565

If there continue to be more bosh-lite specific issues resulting from these values we can revisit this. Thanks!

PR #567, opened 2018-07-19
Enable SpaceDeveloper self-service for network policies

jmcarp:

Required for users to set network policies; without this, container
networking isn't usable. Note: we could drop this property if
https://github.com/cloudfoundry/uaa-release/pull/96 is merged.

Chatted on slack about this with @staylor14

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -945,6 +945,7 @@ instance_groups:
     properties:
       uaa_client_secret: ((uaa_clients_network_policy_secret))
       uaa_ca: ((uaa_ca.certificate))
+      enable_space_developer_self_service: true
       database:
         type: mysql
         username: network_policy


jmcarp:
Alternatively, we could set https://github.com/cloudfoundry/cf-networking-release/blob/master/jobs/policy-server/spec#L52-L54, which seems simpler than adding a new permission to every user. This could go in the base manifest or an opsfile. I think the former makes more sense, because I don't see the purpose of enabling container networking by default if users can't set network policies, but open to suggestions.

cc @rogeruiz

rogeruiz:
I agree, it's weird that we have to explicitly enable this users with `network.write` if they're `SpaceDevelopers`. Is there a real-world use case where we want to disallow `SpaceDevelopers` from setting network policies? cc: @staylor14

heyjcollins:
Hi @jmcarp  - the title is specific to network.write but I see a lot more permissions granted in the changed file.

Can you describe why these permission updates are needed?
We'd like to know generally and will also need to provide info in our release notes.

Thanks,

Josh

rogeruiz:
@heyjcollins I believe those permissions are what we are used to seeing for all users. We don't want to add `network.write` only, we're adding all the permissions we need for users of CF. But, see @jmcarp's  comment above. I don't believe we need this anymore. 

We're now leveraging the changes to the `policy-server` job in `cf-networking-release` to allow `SpaceDevelopers` to set network policies. 

An aside is **why isn't this property default to true** in `cf-deployment` if container networking is also turned on by default in this deployment? it's pretty useless to have container networking on, but not allow users to be able to use it. Is it an oversight? or is there a specific reason?

jmcarp:
@heyjcollins: I updated this patch to set the `enable_space_developer_self_service` property for the policy server instead of the default user authorities. As far as I can tell, if cf-deployment enables container networking by default, we should also allow developers to set network policies--if we don't, why run container networking in the first place?

jmcarp:
Ping @heyjcollins

jmcarp:
Or if this would be considered a breaking change, we can add an ops file.

PR #571, opened 2018-07-24
1803 ops files

ankeesler:

We (the Garden/BOSH Windows teams) are in the process of making Windows Server 2016 version 1803 the default platform for CF Windows components. We want to add ops files to help with deploying 1803 Windows cells.

Currently the 1803 stemcells are not available on BOSH.io which is why we want to keep these ops files in the experimental directory.

Tracker story: https://www.pivotaltracker.com/story/show/158274480

- [ ] YES 
- [X] NO

NOTE: we have run CATs using almost identical versions of these ops files in our CI.

If you have access to an 1803 stemcell and rootfs release, we have provided ops files in the experimental directory that will help you deploy a Windows 1803 cell.

Nope.

- [X] NO

- [ ] GA'd feature/component

CC'ing our anchor, @natalieparellano.


ishustava:
Hey @ankeesler,

How do you feel about renaming this ops file to something more specific since we already have `windows-cell.yml` ops file in the "standard"/non-experimental `operations` folder?

Maybe something like `windows1803-cell.yml`? Although that is inconsistent with `windows2016-cell.yml`. What do you think?

ankeesler:
Hi @ishustava! Would you and your pair have time this afternoon to chat about this question you have posed? We had planned to have this ops file named this way for various reasons, but we definitely understand there are also a number of reasons to name it otherwise as you suggested. I feel like we could benefit from sharing the reason for our naming strategy and getting your opinion on how to move forward.

ankeesler:
For context:
- Iryna and I chatted, and we agreed that this ops file should be renamed to `windows1803-cell.yml`.
- RelInt will check in the rename for this file.

PR #588, opened 2018-08-08
Revert log-cache to 1.3.0

toddboom:

It looks like cf-d 3.0.0 was released with log-cache 1.3.0, but 3.1.0 appears to have automatically grabbed the latest version (1.4.1), which has some known issues. This commit will roll back to the known-good version. We'll remove the packages for both 1.4.0 and 1.4.1 to prevent automatic upgrades until 1.4.2 is released.

- [x] YES 
- [ ] NO

Reverts log-cache to 1.3.0 due to memory management issues under high load.

No.

- [ ] YES
- [x] NO

- [x] GA'd feature/component

/cc @hdub2 @colins

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -2144,9 +2144,9 @@ releases:
   version: "2.0"
   sha1: df91c31c15da68c7b2905645b20b93a930627d46
 - name: log-cache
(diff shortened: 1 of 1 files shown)


ishustava:
Hey @toddboom,

I believe we have reverted the version via our pipeline, and since then have upgraded to log-cache `v1.4.4`. I'm going to close this and delete the branch.

PR #593, opened 2018-08-14
Add ops-file for SMB Volume Service

paulcwarren:

_Describe the change and why it's needed._
As we make our way towards GA the Persi team would like to add an ops file for our SMB Volume Service.

_Include any links to other PRs, stories, slack discussions, etc... that will help establish context._

[#158809215](https://www.pivotaltracker.com/story/show/#158809215)

Signed-off-by: Dave Walter <dwalter@pivotal.io>

- [ ] YES 
- [X] NO

CATs do not test volume services.  Both NFS and SMB Volume Service is tested in [our pipeline](https://persi.ci.cf-app.com).

There is a new ops file that enables the SMB Volume Service.  If applied, SMB Volumes would be available to developers via the `Existing` plan of the `smbservice` from the `AzureFileBroker`.

No

_Does this introduce changes that would require operators to take action in order to deploy without a failure?_

- [X] NO

There is an errand that pushes the broker app but this is obviously short-lived.

- [X] GA'd feature/component

@davewalter @julian-hj @mariash


heyjcollins:
Hey @paulcwarren - we typically accept ops files for new features into the experimental directory first, then promote up to operations/...

Is this actually READY for GA as implied by putting it directly into /operations/...?

If you're truly ready to GA, then it's fine to go as-is, but it's unusual to jump directly to the standard ops files.

We'd be interested in learning how you validate your file system services generally as we're curious if we should be covering this in CATs. Can you hit us up on Slack?

paulcwarren:
On vacation @heyjcollins so responding here as I am sitting at the airport.   

I defer to @julian-hj on the experimental/GA front.   I can see it both ways.  I am not sure which he would prefer though.

On the testing front, we have a set of acceptance tests called PATS that run for every broker/driver combo [including SMB](https://persi.ci.cf-app.com/teams/main/pipelines/persi?groups=cifs).   When we started persi, folks where actually against having anything volume-related in CATs so we rolled our own.  In fact, despite volume services being a feature of its system, Diego doesn't actually test them either.   Instead, they defer to our pipelines.  So it is absolutely a devolved model right now.  But whether it is now time to have something in CATs is a good question.  I think we would be open to having other teams consume our PATs.  We can certainly put some effort into producing some docs in [that repo](https://github.com/cloudfoundry/persi-acceptance-tests) to make them more consumable and putting ops files for supporting components in cf-d's `operations/test` folder.  Again, @julian-hj  will have thoughts on this too.

ishustava:
We spoke with @julian-hj [on slack](https://cloudfoundry.slack.com/archives/C0K0JF319/p1534966585000100) and agreed to move this ops file to the experimental folder.

paulcwarren:
This has been handled manually.

PR #595, opened 2018-08-16
Remove log-cache-scheduler

paolostivanin:
Remove log-cache-scheduler from the scheduler if disable-log-cache ops file is used

- [ ] YES 
- [x] NO (we don't run the full cf-acceptance-tests. We deployed it on 4 different iaases, ran our tests and everything was fine)

- [x] NO

- [x] GA'd feature/component


toddboom:
@paolostivanin Nice catch, thanks!

PR #598, opened 2018-08-17
Ops file for configuring CC to use Log Cache instead of Traffic Controller for app container metrics

tcdowney:

Cloud Controller typically uses Traffic Controller to get container metrics (cpu/disk/memory utilization) for running apps. Loggregator has introduced a new component, Log Cache, that they would like us to use instead. 

This ops file configures CC to use Log Cache instead of Traffic Controller, so we can start kicking the tyres on Log Cache for container metrics.

Here is the story in the CAPI backlog: https://www.pivotaltracker.com/story/show/159042437

- [ ] YES 
- [X] NO

We haven't run CATS against it yet, but we're working on it. 😄 

Adds a new experimental ops-file for using Log Cache to provide container metrics on the CC API.

Nope.

- [X] NO

- [ ] GA'd feature/component

@Gerg @zrob @ericpromislow


staylor14:
Merged manually

PR #608, opened 2018-08-31
Be explicit about which credhub to consume

aeijdenberg:

Make the bosh links explicit for credhub so that this won't fail if another instance of credhub is deployed.

For various reasons we already have a credhub running in our CF deployments. v4.0.0 broke on deployment as it was unable to resolve the bosh link for the new CF credhub since multiple link providers for type credhub existed.

Unsure. We've deployed the changes as described in this PR via an operator file in our environment.

As I understand it, this should not impact operators.

No

No

GA - I think?

Not urgent - we're working around via an operator file.

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -776,6 +776,9 @@ instance_groups:
               name: policy-server
   - name: cloud_controller_ng
     release: capi
+    consumes:
+      credhub:
+        from: cf-credhub
     provides:
       cloud_controller: {as: cloud_controller, shared: true}
     properties:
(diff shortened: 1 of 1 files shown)


aeijdenberg:
Would be nice to get some feedback on this PR...

aeijdenberg:
Hi @heyjcollins - thanks for looking into it and making a reasoned comment before closing it.

It's hard to disagree with your line of reasoning, although I'd throw up the minor point that being explicit is always safer than implicit.

We'll continue with our operator file as a workaround for now.

PR #611, opened 2018-09-04
Remove more references to 2016 in Windows 1803 opsfiles

sesmith177:

- changed the stemcell alias to 1803
- changed the instance group to 1803
- added bosh-dns-alias for windows1803-cell

We are not planning on changing the name of the stack to avoid breaking existing apps

Sorting through the mess of Windows versions (2016 vs. 1709 vs 1803, etc.) can be confusing. In the case here where the stemcell version is important (i.e. 1709 vs 1803) we feel that it is easier to understand what is being deployed if the 1803 version is called out explicitly where posssible

Story: https://www.pivotaltracker.com/story/show/159672477

- [x] YES 
- [ ] NO

Diego cells running Windows 1803 now have instance group name windows1803-cell. Previously, it was windows2016-cell

Yes, it changes the instance group name of the Windows 1803 cell to `windows1803-cell`, although the opsfile is experimental

- [x] NO

- [ ] GA'd feature/component

cc: @ankeesler


heyjcollins:
Hi sesmith177, 
I know folks on your team spoke with Steve Taylor this week and I believe you decided to reconvene and make a decision about how to proceed. I wanted to let you know I pulled this story back into the icebox and I've blocked it on a pending decision from y'all. 

If/When you finalize your approach, could you please reach out to RelInt and let us know so we can remove the blocker and proceed?

Thank you!

Josh

sesmith177:
@heyjcollins @staylor14 After discussing as a team, we feel that having the instance group name, stemcell OS, and the stemcell alias all match (e.g. `windows1803-cell`, `windows1803`, and `windows1803`) is the way to go forward.  This is because it is not possible to just update the stemcell version and deploy -- the releases that are deployed on the cell are different as well. For example, the rootfs `windows2016fs-release` is not compatible with the 1803 stemcell, and the `windows1803-release` is not compatible with the 2016/1709 stemcell.

So while 1709 -> 1803 is technically a semi-annual bump and not an LTS bump, the two OS versions are not compatible from a Windows container / rootfs perspective. So upgrading from one to the next is effectively a major version bump for operators (though not for app developers -- Microsoft claims that any .NET apps compatible with one version remain compatible with next). 

As a side note, we probably should have named the existing `windows2016-cell` instance group `windows1709-cell` at the beginning, but that time has passed.

genevieve:
@natalieparellano Should we close this PR?

davewalter:
Closing stale PR. Please feel free to re-open it if desired.

PR #612, opened 2018-09-05
remove log-cache group-reader job

attack:
Signed-off-by: Mark Gangl <mgangl@pivotal.io>

We're removing the experimental group_reader feature, and so we're dropping it from cf-deployment.

The feature was unstable and addressing a problem for which we have not established much customer support.

Feel free to reach out to us in #cf-log-cache on the cloudfoundry slack. This should just remove one process from the set that runs on the doppler VM.

- [X] YES 
- [ ] NO

Log-cache is removing the experimental group_reader feature

No - it precludes it. `log-cache` v1.4.5 will no longer contain this job, so it needs to be dropped from the manifest before that version bump.

- [X] NO

- [ ] GA'd feature/component

@hdub2 @attack @toddboom @cloudfoundry/cf-log-cache

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -1335,10 +1335,6 @@ instance_groups:
     properties:
       gateway_addr: localhost:8081
     release: log-cache
-  - name: log-cache-group-reader
-    properties:
-      port: 8084
-    release: log-cache
   - consumes:
       reverse_log_proxy: {from: reverse_log_proxy}
     name: log-cache-nozzle
--- a/operations/disable-log-cache.yml
+++ b/operations/disable-log-cache.yml
@@ -4,8 +4,6 @@
 - type: remove
   path: /instance_groups/name=doppler/jobs/name=log-cache-gateway
 - type: remove
-  path: /instance_groups/name=doppler/jobs/name=log-cache-group-reader
-- type: remove
   path: /instance_groups/name=doppler/jobs/name=log-cache-expvar-forwarder
 - type: remove
   path: /instance_groups/name=doppler/jobs/name=log-cache-cf-auth-proxy


ishustava:
Hey @attack,

We consider job removal a breaking change. Here are some examples of breaking changes mentioned in our PR template:

> Examples of breaking changes:
> 
> - changes the name of a job or instance group
> - moves a job to a different instance group
> - deletes a job or instance group
> - changes the name of a credential

As such, we'll have to do it in a major release of cf-deployment. Since [the scope for `v5`](https://docs.google.com/document/d/1M674BR1gHeKJKJ-lctG-jGDe4oYtRaM0ZGeIfQhB6MQ/edit) is already set, I'll queue it up for `v6`. In the meantime, I'll pause log-cache release from being autoupdated in our pipeline.

Also, you've marked that this change is for an experimental feature, but log-cache is in fact GA'd because it's part of the `cf-deployment` base manifest (`cf-deployment.yml`).

hdub2:
Hey @ishustava - while `log-cache` is a GA'd product, the group-reader feature that we are removing is an experimental feature that we've chosen to deprecate. It has been documented as such since July 6th, when its endpoints were moved from `/v1` to `/v1/experimental`.

ishustava:
Thanks for this context @hdub2, that makes sense.

Our PM (@heyjcollins) usually sends out a doc via email with proposed scope for the next major release. After reviewing the email again, I realize that I was wrong about the scope being set (the deadline is this upcoming Monday, September 17th 2018). I'll add this change to the doc and we'll include this change in `v5`. Sorry for the confusion!

To clarify, just submitting a PR is totally fine. If the PR contains a breaking change, depending on the timing, we will include it in either the next major release or the one after. You could also comment or make suggestions on the doc if you'd prefer that.

hdub2:
OK cool - that's great news! I think PRs will continue to be our preferred method for communicating changes. I may have read the 'removal of a job is a breaking change' with a bit of selective comprehension. If it's ok, I may follow up with you directly about how the criteria for 'breaking change' have been defined.

hdub2:
Hiya folks - we cut `log-cache-release` v1.4.5 yesterday, so the requisite version for this PR is now available!

PR #614, opened 2018-09-06
Add an SMB Broker BBR ops-file

davewalter:

We just added a test case fore the experimental SMB volume service and part of that is to add an ops-file here to add the BBR jobs to the `backup-restore` instance.

https://www.pivotaltracker.com/story/show/151415093
BBR PR: https://github.com/cloudfoundry-incubator/disaster-recovery-acceptance-tests/pull/43

- [ ] YES 
- [x] NO

This isn't really applicable since the ops-file we are adding is for BBR. However, it has passed DRATs in our CI.

New Ops-files
- operations/backup-and-restore/enable-backup-restore-smb-broker.yml
  - Enables the backup and restore of SMB service broker instances and bindings. Requires enable-backup-restore.yml.

- [x] NO

- [x] NO

- [ ] GA'd feature/component

I put the new ops-file in the `operations/backup-and-restore` directory. Please feel free to move it to `operations/experimental` if you feel that this is more appropriate.

@julian-hj @paulcwarren @mariash


ishustava:
Thanks for this PR @davewalter.

I was recently reviewing backup-and-restore ops file and noticed that other components are using links for the database properties. Is this something your team is/can considering doing? It's definitely not a blocker for merging this PR though.

davewalter:
Hi @ishustava,

Thanks for taking a look at this. I don't think it's currently on our radar, but I can definitely add a story to our backlog and discuss with @julian-hj @paulcwarren and @mariash.

Regards,
Dave

davewalter:
@julian-hj I think we could have the `nfsbrokerpush` job provide a `nfsbrokerdb` link that the `nfsbroker-bbr` job could consume. That way, if there was a database host configured on the `nfsbrokerpush` job, which I assume would be required for TLS but not otherwise, it would be automatically passed on to the `nfsbroker-bbr` job. The `nfsbroker-bbr` job would also consume the _main_ `database` link that is provided by MySQL/PXC, and it could choose where to get it's database host configuration from, favouring the host from the `nfsbrokerdb` link, if provided, over the address of the first instance in the `database` link, just as the `nfsbrokerpush` template currently does. cc @ishustava

julian-hj:
@davewalter oh, maybe that would work.  A couple questions:
1) can errands provide links?  I thought jobs had to actually run to provide links?  I could definitely be wrong about this though.
2) what would this buy us that we don't already get from just using bosh variables?  Is there a case where the database host is unknown at deployment time?

davewalter:
> 1. can errands provide links?  I thought jobs had to actually run to provide links?  I could definitely be wrong about this though.

That is an excellent question! I'll look into it.

> 2. what would this buy us that we don't already get from just using bosh variables?  Is there a case where the database host is unknown at deployment time?

Personally, I think you are right that, as things have become more templatised and modular with things like credhub and ops-files, this doesn't buy us as much as it used to when you had to maintain your own monolithic manifest, but it still reduces the footprint of the job config that we would have to maintain.

davewalter:
> 1. can errands provide links?  I thought jobs had to actually run to provide links?  I could definitely be wrong about this though.

A [conversation](https://cloudfoundry.slack.com/archives/C02HPPYQ2/p1537379145000100) with the BOSH team has confirmed that you are correct.

PR #634, opened 2018-10-04
Log cache certs reference variables from CA variables

ishustava:

This change switches CA references introduced by #535 to be from their respective CA variables rather than certificate variables.

This reduces log availability downtime during CA rotation.

More information about why CAs need to be referenced this way is described in #305 

N/A

Switch log cache CA references to reduce log availability downtime during CA rotation.

No

- [x] NO

- [x] GA'd feature/component

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -1178,7 +1178,7 @@ instance_groups:
       egress_port: 8080
       health_addr: localhost:6060
       tls:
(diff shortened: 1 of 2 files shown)


jtuchscherer:
cc @toddboom @attack @hdub2 @pianohacker who are working on the log-cache team

PR #637, opened 2018-10-09
Update SMB Volume Service Ops Files

davewalter:

This PR updates the SMB volume service ops-files to replace references to `azurefilebroker` with the more generic `smbbroker`.

We have undertaken a track of work to replace the `azurefilebroker` with a more generic `smbbroker` than can bind existing SMB shares to an app in much the same way that the `nfsbroker` does with existing NFS volumes.

https://www.pivotaltracker.com/story/show/160431033

- [ ] YES 
- [x] NO - Not applicable, but this change has passed our CI.

The enable-smb-volume-service.yml experimental ops-file has been updated to replace the azurefilebroker with a more generic smbbroker than can bind existing SMB shares to an app in much the same way that the nfsbroker does with existing NFS volumes.
The use-pxc-for-smb-volume-service.yml experimental ops-file has been removed as the new smbbroker does not support a SQL backing store.

_Does this introduce changes that would require operators to take action in order to deploy without a failure?_

- [x] YES
- [ ] NO

We have removed the `use-pxc-for-smb-volume-service.yml` ops-file as we are not supporting a SQL backing store for the `smbbroker`.

- [x] NO

- [ ] GA'd feature/component

@julian-hj @paulcwarren @mariash


davewalter:
Please note that this PR requires a version of `smb-volume-release` higher than `0.1.12`.

julian-hj:
I think you will want to pick up smb-volume-release 0.2.0 so that you don't hit errors about a missing smbbrokerpush job when applying this ops file.

ishustava:
Hey @davewalter @julian-hj,

We're going to revert the merge as we've noticed that there are breaking changes in this PR, which we will include in `v6` coming later this month. Sorry all for the confusion!

PR #638, opened 2018-10-10
Use UAA client credentials instead of username/pw for nfsbrokerpush errand

julian-hj:

This change registers a UAA client for the nfsbrokerpush errand to use, and uses those client credentials ranther than the admin username and password to push the nfsbroker app to cf.

Using username/password authentication in non-interactive clients like bosh errands means that it is not possible to turn on 2 factor authentication.  This change helps pave the way for cf operators to enable 2 factor authentication.

[#159187003](https://www.pivotaltracker.com/story/show/159187003)

- [ ] YES 
- [x] NO

(We did test the change through our pipeline however, and run our PATS tests, so we are confident it is tested.)

Maybe not at all?  Once all errands are similarly modified, probably it would make sense to put a generic message in the release notes saying something like "Operators can now enable two factor authentication in CF without breaking errand execution" or something like that.

No.
(It does require nfs-volume-release 1.6.0 or later, but I think cf-d should be picking that up automatically?)

- [x] NO

- [x] GA'd feature/component

@julian-hj @paulcwarren @davewalter @mariash

### DIFF
--- a/operations/experimental/add-cflinuxfs3.yml
+++ b/operations/experimental/add-cflinuxfs3.yml
@@ -2,9 +2,9 @@
   path: /releases/name=cflinuxfs3?
   value:
     name: cflinuxfs3
-    sha1: 7ed10166369721773b54a6ed2af5829271f0af53
-    url: https://bosh.io/d/github.com/cloudfoundry/cflinuxfs3-release?v=0.31.0
-    version: 0.31.0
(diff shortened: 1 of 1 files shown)


julian-hj:
Note--this PR should maybe be blocked on [#161104950](https://www.pivotaltracker.com/n/projects/892938/stories/161104950) as there are still some lingering quirks in the cf cli when spaces are created using client credentials authentication.

PR #644, opened 2018-10-19
Add a manifest to support ubuntu-trusty

xiaozhu36:
This PR aims to provide a manifest to deploy CF deployment using ubuntu-trusty stemcell and then to satisfy our some customer 's using ubuntu-trusty stemcell requirements.


heyjcollins:
Hi @xiaozhu36,

If there are operators who would like to deploy their foundations with ubuntu-trusty, they can create and manage their own custom ops file to do so.
However, now that Xenial is default in cf-deployment and there's no intention to support it (or validate it) going forward, we will not accept PR's to enable it.

Thanks very much,

Josh

PR #646, opened 2018-10-23
Add expvar forwarder job to loggregator-agent addon

bradylove:

The loggregator agent has moved all its metrics to a health endpoint served up by the `expvar` package in the Golang stdlib. This PR adds a job to read the metrics from that endpoint and send them to the loggregator system.

This change requires `loggregator-agent-release` v3.0 which has not yet been cut. We will be cutting that release soon.

We believe that Loggregator component metrics should be accessible when there are issues with reading metrics from the firehose.

* https://www.pivotaltracker.com/story/show/161400481
* https://www.pivotaltracker.com/story/show/161236250

- [X] YES (In our pipelines)
- [ ] NO

Metrics can be retrieved from the Loggregator Agent via HTTP by sending a `GET` request to `localhost:<PPROF_PORT>/debug/vars` from the local VM.

Yes and no, all the metrics should appear in the firehose as they were before, however the expvar forwarder is required to continue receiving those metrics in the firehose.

- [X] NO

- [X] GA'd feature/component

/cc @jtuchscherer @MasslessParticle

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -26,6 +26,71 @@ addons:
           agent:
             cert: "((loggregator_tls_agent.certificate))"
             key: "((loggregator_tls_agent.private_key))"
+  - name: loggr-expvar-forwarder
+    release: loggregator-agent
+    properties:
+      log_agent:
+        ca_cert: "((expvar_forwarder.ca))"
+        client_cert: "((expvar_forwarder.certificate))"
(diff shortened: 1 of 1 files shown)


bradylove:
We have cut v3.0 of the `loggregator-agent-release` and this PR can now be merged and `loggregator-agent-release` can be unpinned from v2.3.

PR #660, opened 2018-11-09
Use 10 reservable ports instead of 100

cdutra:
- this can be expensive in certain IaaSs
- bbl version 7+ will only provision 10 ports for
  TCP Router on AWS

Co-authored-by: Christopher Hunter <hunter.christopher@icloud.com>

Decreases the number of reservable ports to 10.

It can be expensive in some IaaSs to provision 100 ports.
bosh-bootloader version 7+ will only provision 10 ports for TCP Router on AWS

bosh-bootloader is decreasing the number of provisioned reservable ports for AWS.

- [ ] YES 
- [x] NO

TCP Router available reservable ports are now 1024-1033.

If a deployment is using any TCP Router port outside of 1024-1033 range it will break services on those ports.

- [x] NO

- [ ] GA'd feature/component

@cloudfoundry/cf-infrastructure

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -866,7 +866,7 @@ instance_groups:
         router_groups:
         - name: default-tcp
           type: tcp
-          reservable_ports: 1024-1123
+          reservable_ports: 1024-1033
         sqldb:
           host: sql-db.service.cf.internal
           type: mysql


ishustava:
Hey @cdutra & @crhntr,

Thanks for this PR. 

This may be a breaking a change because the `router_groups` property [will not change for upgrade deploys once it's seeded in the database](https://github.com/cloudfoundry/routing-release/blob/ad223c54986a40ec42af16b80830a371819c8e89/jobs/routing-api/spec#L112). This means that operators will need to update this property "manually" via a call to the routing API. Routing API has its own CLI, `rtr`, and it might be a little unusual to ask operators to do this. It would be nice to avoid this change in the base manifest if possible.

You've mentioned that some IaaSes have high costs for provisioning ports. Do you know how many IaaSes are affected?

cdutra:
@ishustava The Azure we [decreased from 150 to 5](https://github.com/pivotal-cf/terraforming-azure/commit/16d3bb5b449245d451e36d500fa431b27989c6e4) because you pay for every single one them.
On AWS we hit a limitation of 50 listeners per Network Load Balancer. I mean, we could still create multiple load balancers and add them to the TCP Router VM.

ishustava:
Ok, thanks for this context @cdutra.

We'll proceed with this as a breaking change and include instructions on how to update reservable ports via routing API. 

cc @heyjcollins

PR #662, opened 2018-11-12
Logging cert dns names

MasslessParticle:

Add instance group dns names to loggregator certs

Loggregator has received feedback that other teams would like the components to use the DNS name as a Common Name because many clients expect this as the default behavior.

https://www.pivotaltracker.com/story/show/161725971

- [x] YES 
- [ ] NO

Loggregator certs now include DNS names as alternative names.

No, but -- the new alternative names won't be present on the certs unless the operator regenerates them

- [x] NO

- [x] GA'd feature/component

@johncornish @jtuchscherer

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -1794,6 +1794,8 @@ variables:
     common_name: statsdinjector
     extended_key_usage:
     - client_auth
+    alternative_names:
+    - uaa.service.cf.internal
 - name: loggregator_tls_agent
   type: certificate
   options:
@@ -1810,6 +1812,8 @@ variables:
     extended_key_usage:
     - client_auth
     - server_auth
+    alternative_names:
+    - doppler.service.cf.internal
 - name: loggregator_tls_tc
   type: certificate
   options:
@@ -1818,6 +1822,8 @@ variables:
     extended_key_usage:
     - client_auth
     - server_auth
+    alternative_names:
+    - log-api.service.cf.internal
 - name: loggregator_tls_cc_tc
   type: certificate
   options:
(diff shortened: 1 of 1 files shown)


heyjcollins:
looks interesting. Please resolve the issues above and we'll consider pulling it in.
Thanks!

MasslessParticle:
@heyjcollins We're a little confused about this one. We're adding this commit directly to develop but our test is failing. I'm not sure where to take it from here?

ishustava:
Hey, @MasslessParticle, 

I've changed the base branch of the PR to `develop`, and now it's happy.

MasslessParticle:
@ishustava ohhhhh, I get it. Thanks for your help!

ishustava:
Could you give us a little more context around this change?

I've looked at the story, and it asks to change **all** loggregator certs to make traffic more secure. It makes sense for one-way TLS. In this case, the server certificate should have the right common name or SAN(s) so that clients can perform hostname verification. 

However, I don't see that it is necessarily more secure in the case of mTLS and client certificates. It doesn't make much sense to add DNS names to client certificates since it's uncommon for servers to perform a similar check for client certificates.

More importantly, the pattern in cf-deployment today seems to be to re-use client certificates on multiple client instances. For example, `statsd_injector` runs on `uaa`, `api`, and `scheduler` instance groups but it uses the same client certificate to connect to loggregator API. Which DNS name is it appropriate to use for this certificate?

MasslessParticle:
We've added the other statsd hostnames.

cdutra:
@MasslessParticle Are you actively working this PR? Our pipeline to guarantee that branches don't stale is failing because this branch is old. Could you please make the changes and make sure they work so we can move forward merging the PR.

MasslessParticle:
@cdutra done, thanks!

PR #664, opened 2018-11-13
Default experimental_containerd_mode to true

Callisto13:

This will default the runtime wrapper in Garden to containerd.

Containerd has been running in PWS for some time and this is part of our plan to fully move this to be the default container lifecycle management.

https://www.pivotaltracker.com/story/show/161584005

- [x] YES 
- [ ] NO

Operators will likely not notice anything unless they are currently using the operations/experimental/use-garden-containerd.yml ops file.

Containerd is not yet compatible with garden bpm or rootless, so containerd will be disabled if either use-bpm-garden.yml or rootless-containers.yml is specified.

No

- [x] NO

- [ ] GA'd feature/component

@BooleanCat

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -1347,10 +1347,13 @@ instance_groups:
           ((application_ca.certificate))
           ((credhub_ca.certificate))
           ((uaa_ca.certificate))
+  - name: containerd
+    release: garden-runc
   - name: garden
     release: garden-runc
     properties:
       garden:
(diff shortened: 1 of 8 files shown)


Callisto13:
@ishustava sorry we neglected to mention something in this PR's description:

It would be good if a stemcell bump could be included in the release that these containerd changes go into. There is no risk of failure without one, but a rolling deploy may result in some (harmless) state being left on disk until the next recreate/stemcell bump.

PR #677, opened 2018-12-11
Make TLS parameter configurable for syslog OPS file

axelaris:

`tls_enabled` configuration parameter was hardcoded to `true`. I made it configurable by setting variable. 

I used `enable-component-syslog.yml` OPS file when deploying CF by Concourse (using cf-concourse-tasks) repo. It was not easy to overwrite hardcoded value, so I decided to make it configurable.

Did not run acceptance tests because it was just change in an OPS file.

I added that variable to example var file so we can treat it as a kind of documentation.

NO

- [ ] NO

- [ ] GA'd feature/component


ishustava:
Hey @axelaris,

Not being able to turn off TLS for component syslog was intentional in cf-deployment. See #477 for more details.

You can disable it by removing the `tls_enabled` property with your own ops file like so:

```
- type: remove
  path: /addons/name=syslog_forwarder/jobs/name=syslog_forwarder/properties/syslog/tls_enabled
```

heyjcollins:
Hey @axelaris,

Thank you for submitting this PR. As Iryna mentioned, the decision to hard-code tls enablement was intentional as it's consistent with our aim to make CF-D secure by default. 

My hope is creating a custom ops file as recommended will work for you and I'm going to go ahead and close this PR.
If you want to continue the dialogue on this, please feel free to comment and reopen.

Best regards,

Josh

PR #696, opened 2019-01-17
update cf login command to specify -u admin

scottschulthess:

The docs specify logging in into bosh-lite post deploy using "admin" username, but cf login prompts for email by default


scottschulthess:
Im a public member of the https://github.com/zipcar which has a corporate CLA and I also think I have done the individual cla

PR #712, opened 2019-02-11
Populate the credhub link properties

paulcwarren:

These properties need to be specified otherwise the credhub link is unusable.

Note, in case you are wondering, because it is required by both the credhub-cli and the credhub Go library to authenticate with UAA, we included the UAA CA certificate.  

As credhub is becoming our default store, we now want to consume the credhub link to populate our brokers (and migrator). 

[#161974894](https://www.pivotaltracker.com/story/show/161974894)

- [ ] YES 
- [X] NO

No existing component (including credhub itself) currently uses the link so running CATS would be a no-op so we didn't do it.

Credhub link properties updated to make bosh link consumable.

No

- [X] NO

- [X] GA'd feature/component

@davewalter @julian-hj

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -1713,7 +1713,6 @@ instance_groups:
             operations: [read]
         ca_certificate: |
           ((credhub_ca.certificate))
-          ((uaa_ca.certificate))
         data_storage:
           database: credhub
           host: sql-db.service.cf.internal
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -1711,6 +1711,9 @@ instance_groups:
           - path: /*
             actors: ["uaa-client:cc_service_key_client"]
             operations: [read]
+        ca_certificate: |
+          ((credhub_ca.certificate))
+          ((uaa_ca.certificate))
         data_storage:
           database: credhub
           host: sql-db.service.cf.internal
@@ -1728,6 +1731,7 @@ instance_groups:
           providers:
           - name: internal-provider
             type: internal
+        internal_url: https://credhub.service.cf.internal
         tls: ((credhub_tls))
     release: credhub
 - name: rotate-cc-database-key


paulcwarren:
Please note that we were able to remove the `uaa_ca_certificate` for our use case so we have removed it from the PR.  Something to think about though is whether, or not, it makes sense to omit it.  Other use cases, for example, a consumer that wanted to run the `credhub-cli` (and not the `credhub` go library as we are) may still need it.

cdutra:
Hi @paulcwarren! Thanks for the PR.

I have few questions before attempting to merge.
1. Is this expected to cause downtime in a HA environment during an upgrade scenario?
2. Have you successfully deployed a manifest with these changes? If yes, how does you deploy command looks like (e.g which ops files do you use)?

davewalter:
Hi @cdutra,

Our understanding from inspecting the [credhub-release job spec file](https://github.com/pivotal-cf/credhub-release/blob/master/jobs/credhub/spec) (specifically, these changes: https://github.com/pivotal-cf/credhub-release/commit/969a3108d757d949759512a800d268f1beb64326 and https://github.com/pivotal-cf/credhub-release/commit/ce4834e9298c2615fbb7bfd739f008dd32b7dc2c) is that credhub does not actually use these properties itself. Instead, these properties were added solely to allow them to be provided to other components via the BOSH link. Given that, we would not expect any downtime if the credhub instance was suitably scaled.

We are deploying with this change in our CI with the following ops-files:
```
operations/use-compiled-releases.yml
operations/scale-to-one-az.yml
operations/windows2016-cell.yml
operations/experimental/use-compiled-releases-windows.yml
operations/enable-nfs-volume-service.yml
operations/test/enable-nfs-test-server.yml
operations/test/enable-smb-test-server.yml
operations/backup-and-restore/enable-backup-restore.yml
operations/backup-and-restore/enable-backup-restore-nfs-broker.yml
operations/experimental/enable-smb-volume-service.yml
operations/update-releases.yml
```
Hope this helps.

Dave

cdutra:
Hi @davewalter, thanks for the reply.
We're running with the following ops files:
```
operations/experimental/fast-deploy-with-downtime-and-danger.yml
operations/use-compiled-releases.yml
operations/rename-network-and-deployment.yml
operations/set-bbs-active-key.yml
operations/test/add-persistent-isolation-segment-diego-cell.yml
operations/test/add-persistent-isolation-segment-router.yml
operations/rename-isolation-segment-network.yml
operations/addons/enable-component-syslog.yml
operations/use-postgres.yml
operations/experimental/enable-tls-cloud-controller-postgres.yml
operations/change-postgres-max-connections.yml
```
and we're getting the following error:
```
Summarizing 1 Failure:

[Fail] [docker] Docker App Lifecycle CredHub Integration when CredHub is configured [AfterEach] service bindings in assisted mode the app should see the service creds
/Users/pivotal/go/src/github.com/cloudfoundry/cf-acceptance-tests/docker/credhub_enabled.go:81
...
[2019-02-21 00:52:01.46 (UTC)]> cf purge-service-offering CATS-6-SERVICE-NAME-ba9121bc5f56f65d
WARNING: This operation assumes that the service broker responsible for this service offering is no longer available, and all service instances have been deleted, leaving orphan records in Cloud Foundry's database. All knowledge of the service will be removed from Cloud Foundry, including service instances and service bindings. No attempt will be made to contact the service broker; running this command without destroying the service broker will cause orphan service instances. After running this command you may want to run either delete-service-auth-token or delete-service-broker to complete the cleanup.

Really purge service offering CATS-6-SERVICE-NAME-ba9121bc5f56f65d from Cloud Foundry? [yN]: EOF
FAILED
```

davewalter:
Hi @cdutra,

It looks like the test should be specifying the `-f` flag to `cf purge-service-offering` to eliminate the confirmation prompt. A [quick search](https://github.com/cloudfoundry/cf-acceptance-tests/search?q=purge-service-offering&unscoped_q=purge-service-offering) shows that everywhere else that the `cf purge-service-offering` command is invoked in CATs includes that flag.

Regards,
Dave

cdutra:
Indeed @davewalter, It's interesting that it was working before. I made the change to add the `-f` flag, tests went green and I'm merging it now. Thank you again for the PR!

PR #741, opened 2019-03-26
Scheduler instance group deploys immediately after api

selzoc:

yup

Moving the `scheduler` instance group to immediately after the `cc-worker` instance group

The cloud_controller_ng job may run migrations against ccdb when it is deployed. This means that for a time the cloud_controller_clock and cloud_controller_worker are running older code against this newly migrated ccdb. This is typically fine, but if those jobs are stopped and restarted for any reason (monit restarts, someone inadvertently running bbr during the deploy), then they cannot come up successfully because their older code is out of date with the db.

To minimize the time between the cloud_controller_ng job deploying and the scheduler (with cloud_controller_clock job) deploying we should have them deploy immediately after cloud_controller.

- [ ] YES 
- [x] NO

Change deploy order of scheduler instance group to minimize time between deploys of `cloud_controller_ng` and `cloud_controller_clock` jobs.

We don't believe so

- [x] NO

no

@tcdowney @Gerg @coreyti

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -1087,70 +1087,6 @@ instance_groups:
             secret: "((uaa_clients_cc-service-dashboards_secret))"
           cc_routing:
             secret: "((uaa_clients_cc-routing_secret))"
-- name: router
-  azs:
-  - z1
-  - z2
-  instances: 2
-  vm_type: minimal
-  vm_extensions:
-  - cf-router-network-properties
-  stemcell: default
-  update:
-    serial: true
-  networks:
-  - name: default
-  jobs:
-  - name: gorouter
-    release: routing
-    properties:
-      router:
-        enable_ssl: true
-        tls_pem:
-        - cert_chain: "((router_ssl.certificate))"
-          private_key: "((router_ssl.private_key))"
-        ca_certs: |
-          ((application_ca.certificate))
-          ((service_cf_internal_ca.certificate))
-        backends:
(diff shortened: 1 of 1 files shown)


cdutra:
Hi @selzoc @tcdowney, thanks for the PR!

That means that `cloud_controller_clock`  has higher priority then `gorouter` and `tcp_router` jobs. Did you run upgrade deployments from the previous setup to this new one? Did you experience any downtime or do you think this might introduce downtime on upgrade scenarios?

tcdowney:
@adobley @aaronshurley 

Would you anticipate this change causing any issues with the routers? From our perspective, the `cloud_controller_clock` _should_ have higher priority than the routers since it relies on the same CCDB as the `cloud_controller_ng` (api) and `cloud_controller_worker` jobs, so when the canary api job runs the latest migrations this creates a window of time where the clock is running older code against a newer CCDB. This is generally ok, but we'd like to minimize this time period because if anything causes the clock to crash it won't be able to restart (it will detect that the database schema is newer than what it expects) and this can lead to CC and Diego getting out of sync.

We believe this change is safe, but don't have the routing expertise to answer @cdutra's question for sure.

cdutra:
@tcdowney Thank you so much for the quick feedback.
 I believe this should be fine. I'll soon merge this PR and if it goes into release candidate it means it went through our pipeline's validations. If I see any issues I'll let you know.

Thanks again!

PR #744, opened 2019-04-01
Our licenses should state org correctly for copyright purposes

johannaratliff:
Signed-off-by: Johanna Ratliff <josmith@pivotal.io>

I was fixing the license for one of our sub projects and realized we don't state org, we just use template in a lot of these.
Year was chosen based off of original license add date.

Fix org name in license

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -187,7 +187,8 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] [name of copyright owner]
+   Copyright 2017 CloudFoundry.org Foundation, Inc.
+
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.


cdutra:
Hi @johannaratliff, thanks for the PR! Can you make sure all authors have signed the CLA, please?

emalm:
Hey, @johannaratliff, the cf-deployment [NOTICE file](https://github.com/cloudfoundry/cf-deployment/blob/master/NOTICE) has this copyright notice filled in correctly. I don't think the copy of the Apache 2.0 License in the repository also needs it filled in, since it's directing readers of the license about how to apply it to their own works.

josephgee:
@cdutra @johannaratliff I believe I'm covered under the corporate CLA and that I've publicized the org memberships. Let me know if I missed something.

cdutra:
@johannaratliff @josephgee Based on @emalm comment I believe this is not necessary. We'll close it, feel free to reopen if you think this is necessary.

cc @davewalter

PR #745, opened 2019-04-01
added operations file to disable route integrity in diego rep

mariash:
[#161142258](https://www.pivotaltracker.com/story/show/161142258)

Co-authored-by: Maria Shaldibina <mshaldibina@pivotal.io>

### Is this a PR to [the develop branch](https://github.com/cloudfoundry/cf-deployment/tree/develop) of cf-deployment?

Yes.

### WHAT is this change about?

Adding an ops file to disable routing integrity, which is turned off by default currently, but is about to be turned on by default.

### WHY is this change being made (What problem is being addressed)?

We'd like to give cf-deployment operators an opt-out configuration for disabling route integrity before we make it the default configuration in a subsequent major version.

### Please provide contextual information.

https://www.pivotaltracker.com/story/show/161142258

### Has a cf-deployment including this change passed our [cf-acceptance-tests](https://github.com/cloudfoundry/cf-acceptance-tests)?

- [ ] YES 
- [X] NO

### How should this change be described in cf-deployment release notes?

_Something brief that conveys the change and is written with the Operator audience in mind. 
See [previous release notes](https://github.com/cloudfoundry/cf-deployment/releases) for examples._

New Ops-files

disable-routing-integrity.yml - Disables container proxy on the Diego Cell `rep`

### Does this PR introduce a breaking change? 

No.

### Will this change increase the VM footprint of cf-deployment?

- [ ] YES --- does it really have to?
- [X] NO

### Does this PR make a change to an experimental or GA'd feature/component?

- [X] GA'd feature/component

### What is the level of urgency for publishing this change?

- [ ] **Urgent** - unblocks current or future work
- [X] **Slightly Less than Urgent**

@cf-diego

PR #746, opened 2019-04-01
inlined enable-routing-integrity.yml into cf-deployment.yml

n4wei:
[#161142258](https://www.pivotaltracker.com/story/show/161142258)

Co-authored-by: Maria Shaldibina <mshaldibina@pivotal.io>

Yes

This is a breaking change that makes route integrity default in cf-deployment intended for the next major version of cf-d. This other PR https://github.com/cloudfoundry/cf-deployment/pull/745 is related and allows operator to opt out of routing integrity via a new operations file.

We would like to make app traffic more secure by turning route integrity on by default.

Tracker story: https://www.pivotaltracker.com/story/show/161142258
PR to opt out of route integrity: https://github.com/cloudfoundry/cf-deployment/pull/745
Subsequent PR to remove the deprecated ops file: https://github.com/cloudfoundry/cf-deployment/pull/753

- [ ] YES 
- [X] NO

By default, enables container proxy on the Diego Cell rep and configures gorouter to opt into TLS-enabled connections to the backend.

Yes. It is intended for the next major version of cf-d.
- changes cf-deployment.yml

- [X] NO

- [X] GA'd feature/component

@cf-diego


Syerram:
@n4wei Deprecation of ops files is two part process. You can find the flow [here](https://github.com/cloudfoundry/cf-deployment/blob/master/ops-file-promotion-workflow.md). 

Basically, you replace the entire ops file with one word `#DEPRECATED` and PR the change. We will release that as a minor release with the deprecation notice that we will remove this ops file in the next major release. 

Also, does the change in `cf-deployment.yml` introduce breaking change?

sunjayBhatia:
@Syerram this is a significant change we would expect to be desirable to make in the next *major* version of cf-d

mariash:
We updated the PR with `#DEPRECATED` comment and left the ops file. We made another PR to remove the ops file in subsequent step: https://github.com/cloudfoundry/cf-deployment/pull/753

mariash:
@cf-rel-int-status-bot I think the test needs to be updated since there is a change to cf-deployment?

davewalter:
Hi @mariash,

Could you please include the fix to the tests in the PR? Also, since this is a breaking change, I think we won't be able to pull it in until v9 at this point.

Thanks,
Dave

CC @Syerram

PR #750, opened 2019-04-05
Merge forwarder agent

chentom88:

Yes

This is change is to move the loggregator forwarder agent from an experimental ops-file into cf-deployment.yml

This change is being made so that forwarder agent can become a proper component of CF

Project: https://www.pivotaltracker.com/story/show/164614284

- [X] YES 
- [ ] NO

This change introduces the forwarder agent component of Loggregator that is the first step into the new simplified Loggregator architecture. 

No

We believe we have followed the Ops-file promotion workflow.

- [X] NO

- [ ] GA'd feature/component

PM: @jtuchscherer
Dev: @chentom88


cf-deployment.yml
```diff
@@ -35,6 +37,12 @@ addons:
         client_key: "((expvar_forwarder.private_key))"
       default_source_id: metron
       counters:
+      - addr: http://127.0.0.1:14823/debug/vars
```
cdutra: Is this the right port, or should be 14824 like the other ones?

jtuchscherer: cc @chentom88 @MasslessParticle

chentom88: @cdutra Good catch, looks like we had it wrong in our original ops file and it was copied over 

https://github.com/cloudfoundry/cf-deployment/blob/develop/operations/experimental/deploy-forwarder-agent.yml#L64

We'll fix it. Thanks.

chentom88: @cdutra Actually that port is correct. We want this metric to be coming from the new forwarder agent which is exposed at port `14823`.

johncornish: This is the right port; loggregator_agent and forwarder_agent use different ports for metrics because they both spin up an http server to host their metrics and are colocated

PR #751, opened 2019-04-05
Add experimental option to have a metric store in CF

johannaratliff:
[#164122999]

Signed-off-by: Johanna Ratliff <josmith@pivotal.io>

Add Metric Store as an option within CF.

Persistence and queryability (PromQL) of metrics downstream from Loggregator.

- [x] YES 
- [ ] NO

Add Metric Store - persistent storage for metrics downstream from Loggregator - as an experimental ops file.

N

- [x] YES --- does it really have to? Yes. But it's an experimental ops file.
- [ ] NO

- [ ] GA'd feature/component

/cc @toddboom @jtuchscherer


cdutra:
Hi @johannaratliff, thanks for the PR!
Just for curiosity, is the plan for this feature to become default in cf-deployment?

johannaratliff:
@cdutra Unclear to me. Consult with @jtuchscherer @toddboom

jtuchscherer:
@cdutra There is a strong case for it to end up as a default in cf-deployment as we are planning to take away the `/query` endpoint from log-cache in favor of providing it through the metric-store. But we are unclear on timing for this part of the work.

PR #754, opened 2019-04-15
Deprecate logging isolation segments

heycait:

_Only PR's to develop are accepted._

Yes

_Describe the change._
Deprecate logging isolation segment ops-files

_Understanding why this change is being made is fantastically helpful. Please do tell..._

Isolated logging was only ever experimental. We no longer support them as they exist in these ops files.

_Include any links to other PRs, stories, slack discussions, etc... that will help establish context._
https://www.pivotaltracker.com/story/show/165038981

- [x] YES
- [ ] NO

Deprecate experimental isolated logging functionality.

_Does this introduce changes that would require operators to take action in order to deploy without a failure? Please see below for few examples._
This PR will break anyone relying on deprecated ops files

- requires out-of-band manual intervention on the part of the operator
- removes Ops-file (e.g. from `./operations/` or `./operations/experimental` folders)

- [ ] YES - please specify
- [x] NO

- [x] NO

- [ ] GA'd feature/component

@MasslessParticle @jtuchscherer

### DIFF
--- a/operations/test/add-persistent-isolation-segment-infrastructure-metrics.yml
+++ b/operations/test/add-persistent-isolation-segment-infrastructure-metrics.yml
@@ -1,3 +1,4 @@
+# DEPRECATED
 - type: remove
   path: /addons/name=prom_scraper
 
--- a/operations/test/add-persistent-isolation-segment-logging-system.yml
+++ b/operations/test/add-persistent-isolation-segment-logging-system.yml
@@ -1,3 +1,4 @@
+# DEPRECATED
 ---
 - type: replace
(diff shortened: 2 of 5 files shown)


Syerram:
@heycait Since this a breaking change, I recommend you
1. Do NOT empty the file but add `#DEPRECATED` as the first line in the file.
2. In our next release, I will add deprecation notice and it's removal in the following major release.

This gives operators enough time to update their deployments.

heycait:
We've updated the PR to only add #DEPRECATED instead of removing the file contents

PR #783, opened 2019-06-26
Add an ops-file to enable TLS on File Server

mariash:

YES.

This is a new ops-file that enables TLS mode on Diego File Server. File Server used to only support plain HTTP mode. Starting with Diego v2.33.0 it can be run in TLS mode.

Adds a security to application staging/starting process. File Server is used for storing application lifecycle binaries.

https://www.pivotaltracker.com/story/show/165872883
https://www.pivotaltracker.com/story/show/166882706

- [X] YES
- [ ] NO

New Ops-Files

Ops-File: operations/enable-tls-on-file-server.yml
Purpose: Enables TLS mode on Diego File Server
Notes: Requires Diego release v2.33.0+

No.

(If deployed with Diego release earlier than 2.33.0 these properties would just be ignored)

- [X] YES - please specify
- [ ] NO

operations/enable-tls-on-file-server.yml

- [X] NO

- [X] GA'd feature/component

@cf-diego


Syerram:
@mariash Is the intent to deprecate http only at some point and have everything over TLS?

sunjayBhatia:
@Syerram eventually yes, but we cannot do so immediately b/c there are existing app definitions that include HTTP URLs and CC is not yet generating HTTPS URLs (eventually Capi will generate only HTTPS URLs but there still may be existing apps running that reference HTTP, so we do have to be careful about disabling the HTTP server)

PR #786, opened 2019-07-02
Disable uaa legacy redirect uri matching

shamus:

Yes.

This PR introduces an experimental ops file which disables the UAA's legacy redirect matching in favor of an exact matching strategy. This exact matching strategy is more in line with the intentions of the OAuth spec.

For inquiring minds, the legacy matching behavior allowed clients to register a URI which could be expanded to match classes of urls.  For example:

- registering `http://example.com/` would match requested redirect uris of `http://example.com/path`, `http://subdomain.example.com`, and more!
- registering `http://example.com/**` will match subpaths of any depth. i.e. `http://example.com/foo`, `http://example.com/foo/bar`

If you're wondering what all this redirect uri talk is about:

https://www.oauth.com/oauth2-servers/redirect-uris/

After applying this ops file, redirect uris will be subject to exact matching.

- registering `http://example.com/` will only match a requested redirect uri of `http://example.com`

While this legacy matching mode is convenient, it allows surprising and unexpected matches.  In general, the UAA team believes being explicit when configuring this value is the best practice. 

- [x] YES
- [ ] NO

We ran CATs with this config:

```json
{
  "api": "api.uaa-test-env.cf-app.com",
  "apps_domain": "uaa-test-env.cf-app.com",
  "admin_user": "admin",
  "admin_password": "<redacted>",
  "skip_ssl_validation": true,
  "include_internet_dependent": false,
  "include_services": true,
  "include_sso": true,
  "include_v3": false,
  "use_http": true,
  "cf_push_timeout": 600,
  "default_timeout": 600,
  "cf_dial_timeout_in_seconds": 30,
  "backend": "diego",
  "enable_color": true,
  "use_existing_user": false,
  "keep_user_at_suite_end": false,
  "artifacts_directory": "/var/vcap/sys/log/acceptance_tests/",
  "credhub_secret": "<redacted>",
  "credhub_client": "credhub_admin_client"
}
```

@wc22222 will provide this

No.  We're providing an experimental ops file that an operator can opt into.  

In order for operators to use this ops file, they will need to independently verify that no users of their foundations have registered clients with the UAA that take advantage of the legacy matching behavior. Apps which rely on clients that take advantage of the legacy matching mode will encounter errors when trying to retrieve tokens from the UAA.

- [ ] YES - please specify
- [x] NO

- [x] NO

- [ ] GA'd feature/component

PMs: @wc22222 @dbeneke
Developers: @shamus @birdrock @joshuatcasey @andrewedstrom


davewalter:
Hi @shamus,

Could you please remove the references to UAA release from the ops-file? We're not planning to merge this until after the cf-deployment v10 work, which includes unpinning UAA to consume v73 in the base manifest.

Thanks,
Dave and @Changdrew 

cc @wc22222 @dbeneke @Birdrock @joshuatcasey @andrewedstrom

shamus:
@davewalter when might this be merged?  :)

PR #788, opened 2019-07-08
Update use-alicloud-oss-blobstore.yml

ninja-at-work:
Enable usage of multiple buckets for fog aliyun

Yes

Enables fog aliyun to operate with multiple buckets as it has already done for bits aliyun.

Bits service for AliCloud utilizes multiple buckets to persist the files. This approach is totally different from fog configuration, which accepts a single bucket only to persist all kinds of resources. This PR aligns fog configuration with bits accepted configuration to enable a smooth transition from bits to fog adapter.
The reason for that is because the bits service is going to be in maintenance mode and deprecated from CF.

Fog configuration:
https://github.com/cloudfoundry/cf-deployment/blob/master/operations/use-alicloud-oss-blobstore.yml#L12

Bits cofiguration:
https://github.com/cloudfoundry/cf-deployment/blob/master/operations/bits-service/configure-bits-service-alicloud-oss.yml#L5

- [ ] YES
- [ ] NO

Yes, the old fog configuration may stop work.

- [ ] YES - please specify
- [X] NO

- [X] NO

- [X] GA'd feature/component

### DIFF
--- a/operations/use-alicloud-oss-blobstore.yml
+++ b/operations/use-alicloud-oss-blobstore.yml
@@ -4,53 +4,68 @@
 - type: replace
   path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/cc/buildpacks/fog_connection
   error: "Please apply 'use-external-blobstore.yml' before applying 'use-alicloud-oss-blobstore.yml'."
-  value: &blobstore-properties
+  value: &buildpacks-blobstore-properties
     provider: aliyun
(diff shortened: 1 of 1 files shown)


acosta11:
Hi ninja-at-work,

With two PR's addressing the same functionality, the team elected to pursue #800 in the interest of avoiding a breaking change to the existing alicloud blobstore opsfile.

Thanks,
Andrew

PR #791, opened 2019-07-12
Internal Markdown links resolve correctly on GitHub

cunnie:
When `deployment-guide.md` was moved below `/texts`, many of the relative links were broken when the markdown was rendered on GitHub, returning 404s when clicked.

This corrects the links so they render properly on GitHub. I chose to use absolute paths (i.e. begin with a "/"), but a relative path would have worked as well.

Two links are updated to use HTTPS instead of HTTP though this is more for consistency than security concerns.

Yes

Fix broken documentation links.

Documentation links were broken; now they are fixed.

- [ ] YES
- [ ] NO
- [X] NOT SURE

Fix broken documentation links.

No.

- requires out-of-band manual intervention on the part of the operator
- removes Ops-file (e.g. from `./operations/` or `./operations/experimental` folders)

- [ ] YES - please specify
- [x] NO

- [x] NO

- [ ] GA'd feature/component

### DIFF
--- a/texts/deployment-guide.md
+++ b/texts/deployment-guide.md
@@ -18,7 +18,7 @@ and installing a BOSH Director.
  
 #### bosh-lite
 If you're deploying bosh-lite to a VM on AWS, GCP, or Azure,
-look at [this guide](iaas-support/bosh-lite/README.md).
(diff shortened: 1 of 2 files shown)


vitreuz:
Looks good! Thanks @cunnie!

PR #807, opened 2019-08-06
add examples of breaking changes.

Syerram:

- Add examples of breaking changes to the versioning doc. 
- Update the PR template to include an example of a property change

Educate component teams on breaking changes.


davewalter:
@Syerram Can you please switch the base branch from `master` to `develop`?

PR #812, opened 2019-08-16
ops: Remove "windows2016" stack

Cara-Sciorilli:
This "windows2016" stack is no longer supported in favor of the "windows" stack. This is a 
name change to remove any confusion about the version of windows being used. 
See https://docs.cloudfoundry.org/devguide/deploy-apps/windows-stacks.html

[#166852962](https://www.pivotaltracker.com/story/show/166852962)

https://www.pivotaltracker.com/story/show/166852962
https://cloudfoundry.slack.com/archives/C0FAEKGUQ/p1565967749085100

- [X] YES
- [ ] NO

### Does this PR introduce a breaking change? Please see definition of breaking change below

- [X] YES - applications using the older unsupported "windows2016" stack would stop working.
- [ ] NO

- [ ] YES - please specify
- [X] NO

@arjun024 @yaelharel @jro-pv


Cara-Sciorilli:
Hi @davewalter,

The `windows2016` stack and `windows2016-cell` have have no connection. The `windows2016-cell.yml` is the ops-file to add the Windows Server 1709 operating system version cell which is also called Windows Server 2016. The `windows2016` stack was originally for the Windows Server 2016 which introduced containers. Since then, later Windows operating systems have been using the same stack, which is the cause for the rename to just `windows` to avoid confusion. Now we are removing the legacy `windows2016` stack.

Cara

davewalter:
Thanks for the clarification @Cara-Sciorilli. I was under the impression that Windows 2016 was being sun-setted due to it no longer being supported by Microsoft. Was I mistaken, or is this just not happening yet?

Regards,
Dave

arjun024:
Yes, Windows Server 2016 has to be retired. But we would like to not conflate the two tasks - the stack and the first OS version that this stack supports happen to have very identical names. For sometime, we have allowed users to use both `windows` and `windows2016` to refer to the exact same thing. This PR removes the stack name `windows2016` which is confusing and has that particular name because of historical reasons. When we're all done with this, we should get to deprecate the older Windows 2016 server support, but yes they are unrelated.

davewalter:
This has been reverted by https://github.com/cloudfoundry/cf-deployment/commit/c0d481f7cc3d3162cd091f67f2f474b5247c7b31

PR #817, opened 2019-08-21
add metrics discovery registrar addon as experimental ops files

kitemongerer:
Signed-off-by: Max Eshleman <meshleman@pivotal.io>

This change adds an experimental ops file to add metrics discovery registrar to VMs. Metrics discovery registrar is a component that registers scrapable metrics endpoints on the VM to NATs.

### What customer problem is being addressed? Use customer persona to define the problem 
This change enables customers and component teams to discover scrapable metrics endpoints across their foundations. After this change, any interested component can subscribe to the NATS queue and discover metrics endpoints.

README explaining the new product:
https://github.com/cloudfoundry/metrics-discovery-release/blob/develop/README.md

- [x] YES
- [ ] NO

### Does this PR introduce a breaking change? Please see definition of breaking change below

- [ ] YES - please specify
- [x] NO

Add ops files that will put the metrics discovery registrar job on all VMs to register scrape configurations with NATS. This allows any external metrics scraper to easily discover scrape endpoints on the platform

- [x] YES - metrics-discovery-release which contains metrics discovery registrar
- [ ] NO

- [x] NO

- [ ] GA'd feature/component

@MasslessParticle @AllenDuet

### DIFF
--- a/operations/experimental/README.md
+++ b/operations/experimental/README.md
@@ -19,6 +19,8 @@ This is the README for Experimental Ops-files. To learn more about `cf-deploymen
 | [`add-deployment-updater-postgres.yml`](add-deployment-updater-postgres.yml) |  **"DEPRECATED because it is inlined in to `cf-deployment.yml`"** | Requires `add-deployment-updater.yml` | **NO** |
 | [`add-deployment-updater.yml`](add-deployment-updater.yml) |  **"DEPRECATED because it is inlined in to `cf-deployment.yml`"** | | **NO** |
 | [`add-metric-store.yml`](add-metric-store.yml) |  **Add Metric Store node for persistence of metrics from Loggregator** | | **NO** |
+| [`add-metrics-discovery-registrar.yml`](add-metrics-discovery-registrar.yml) |  Add metrics discovery registrar to all vms to register scrape configurations with NATS | | **NO** |
+| [`add-metrics-discovery-registrar-windows1803.yml`](add-metrics-discovery-registrar-windows1803.yml) |  Add metrics discovery registrar to windows1803 Diego cells to register scrape configurations with NATS | Requires `../windows1803-cell.yml`, `add-metrics-discovery-registrar.yml` | **NO** |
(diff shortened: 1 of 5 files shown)


davewalter:
Hi @kitemongerer,

We noticed that the main ops-file adds a new `metrics-discovery` release, but that release is not currently available on bosh.io. We would prefer to be able to specify a url/version/sha for the new release in the ops-file so that users don't have to manually upload the release to their BOSH director before deploying. We would also prefer to be able to add a job to our CI to keep this release version up to date as new versions are created. Is there a story we can track so that we know when this is available?

We were also wondering why you created a separate ops-file to add the job to windows1803 cells (`add-metrics-discovery-registrar-windows1803.yml`), instead of a second add-on? It is also worth noting that there are other flavours of windows cells (2012R2, 2016 and 2019) that you aren't including coverage for. Is this intentional?

Regards,
Dave and @vitreuz

davewalter:
Hi @kitemongerer,

Just wanted to check in on this PR to avoid it getting too stale. Please see our comments above that are preventing it from being merged. Also, it looks like it needs to be rebased against develop.

Thanks,
Dave

kitemongerer:
@davewalter I will add a story to rebase this PR. Is there a good example of agents being added to windows VMs that we can follow? The other loggregator agents are added only to 1803 and not as an addon so we just followed that pattern. We also only test on 1803.

davewalter:
Hi @kitemongerer,

> Is there a good example of agents being added to windows VMs that we can follow? The other loggregator agents are added only to 1803 and not as an addon so we just followed that pattern. We also only test on 1803.

If you want to switch to an addon, you can use the existing one you have for linux VMs as a template. You simply need to change the `include.stemcell` criteria from ubuntu stemcells to windows ones. Is it your intention to only support 1803, or should the `metrics-discovery-registrar-windows` job support all versions of Windows? If the former, then I think what you have is fine. 

Regards,
Dave

davewalter:
Hi @kitemongerer,

Just wanted to bump this in your "inbox" to see how you wanted to proceed. Also, it looks like there are now some conflicts that you will need to resolve before we can merge.

Regards,
Dave

heycait:
Hi @davewalter we're going to bump metrics-discovery-release to 2.0.0 in another PR

PR #825, opened 2019-09-17
Add request for acceptance criteria to the PR template

Syerram:
Often times, I (PM) am unable to verify the changes in the acceptance environment. Because of the missing AC, either we accept the change in good faith or pair with devs to find the closest AC (which may or may not be true AC).

With this change, PR authors provide AC for their changes. The AC should be identical to AC they used in their own CI environment. In addition, it forces teams to think ahead on how-to verify changes using cli (or cats).

Furthermore, I removed redundant questions and questions that can be inferred by looking at the PR itself. E.g. ` Does this PR introduce a new BOSH release into the base cf-deployment.yml manifest or any ops-files` can be inferred by looking at the PR itself. There is no further action required by the author (On the contrary, RelInt will need to add the bosh-release to update releases etc.)

### DIFF
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -19,18 +19,19 @@
 - [ ] YES
 - [ ] NO
 
-### Does this PR introduce a breaking change? Please see definition of breaking change below.
+### Does this PR introduce a breaking change? Please take a moment to read through the examples before answering the question.
 
-- [ ] YES - please specify
+- [ ] YES - please choose the category from below. Feel free to provide additional details.
 - [ ] NO
 
 > **Types of breaking changes:**
-> - **causes app or operator downtime**
-> - modifies, deletes or moves the name of a job or instance group in the main manifest
-> - modifies the name or deletes a property of a job or instance group in the main manifest
-> - changes the name of credentials in the main manifest
-> - requires out-of-band manual intervention on the part of the operator 
-> - modifies the ops-file path, changes the type, changes the values or removes ops-files from the followning folders
+> 1. **causes app or operator downtime**
+> 2. increases VM footprint of cf-deployment - e.g. new jobs, new add ons, increases # of instances etc.
+> 3. modifies, deletes or moves the name of a job or instance group in the main manifest
+> 4. modifies the name or deletes a property of a job or instance group in the main manifest
+> 5. changes the name of credentials in the main manifest
+> 6. requires out-of-band manual intervention on the part of the operator 
+> 7. modifies the ops-file path, changes the type, changes the values or removes ops-files from the followning folders
 >    - `./operations/` or `./operations/experimental` 
 >    - `./addons`
 >    - `./backup-and-restore/`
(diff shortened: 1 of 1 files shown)


PULL_REQUEST_TEMPLATE.md
```diff
@@ -45,20 +41,14 @@
...
-- [ ] YES - please specify
-- [ ] NO
-
-### Will this change increase the VM footprint of cf-deployment?
-
-- [ ] YES - does it really have to?
-- [ ] NO
+### Please provide Acceptance Criteria for this change?
 
-### Does this PR make a change to an experimental or GA'd feature/component?
```
Syerram: Please let me know if this question adds value to our PR workflow?

PULL_REQUEST_TEMPLATE.md
```diff
@@ -14,18 +14,14 @@
 
 > _Include any links to other PRs, stories, slack discussions, etc... that will help establish context._
 
-### Has a cf-deployment including this change passed [cf-acceptance-tests](https://github.com/cloudfoundry/cf-acceptance-tests)?
```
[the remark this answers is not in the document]
weymanf: Yes I also think this section forces teams to at least double check that they ran CATs.

PULL_REQUEST_TEMPLATE.md
```diff
@@ -45,20 +41,14 @@
 
 > _Something brief that conveys the change and is written with the **persona (Alana, Cody...)** in mind. See [previous release notes](https://github.com/cloudfoundry/cf-deployment/releases) for examples._
 
-### Does this PR introduce a new BOSH release into the base cf-deployment.yml manifest or any ops-files?
-
-- [ ] YES - please specify
-- [ ] NO
-
-### Will this change increase the VM footprint of cf-deployment?
```
weymanf: We want to keep this section because there might be some non-obvious component changes that require an increase to VMs for it function properly. 

For Example, an extra addon or a new bosh job will require bigger diego cells.

acosta11:
> I like the new question you want to add.
> 
> For the questions that you want to remove, I can see the argument that the information is available in the PR itself, but I think that, historically, we have found that engineers missed key points when reviewing/merging PRs and so we added these questions to draw everyone's attention to these important pieces of information, including that of the people submitting the change.
> 
> Dave

The team discussed this and would like to move forward by starting with the addition of the Acceptance Criteria section for now, and then removing the other questions in a separate PR.

PULL_REQUEST_TEMPLATE.md
Syerram: Interesting. Wouldn't that be a breaking change, because it adds a new job to CF (and thus increases vm footprint)?

Syerram: I also moved this to the breaking change section. See above

PULL_REQUEST_TEMPLATE.md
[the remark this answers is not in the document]
Syerram: @weymanf I ask this question in AC section but maybe it warrants it's own section.

PR #826, opened 2019-09-17
add drain_ca_cert to use-logcache-syslog-ingress ops files

kitemongerer:
Signed-off-by: Caitlyn Yu <cyu@pivotal.io>

This change allows the syslog agent to securely communicate with the log-cache syslog server.

The user is now able to use syslog-log-cache without disabling tls validation

- [x] YES
- [ ] NO

### Does this PR introduce a breaking change? Please see definition of breaking change below.

- [ ] YES - please specify
- [x] NO

Update use-logcache-syslog-ingress ops files to enable secure communication between the syslog agent and the log cache syslog server.

- [ ] YES - please specify
- [x] NO

- [x] NO

- [ ] GA'd feature/component

@MasslessParticle @pianohacker @mjseaman

### DIFF
--- a/operations/experimental/use-logcache-syslog-ingress-windows1803.yml
+++ b/operations/experimental/use-logcache-syslog-ingress-windows1803.yml
@@ -2,3 +2,7 @@
 - type: replace
   path: /instance_groups/name=windows1803-cell/jobs/name=loggr-syslog-agent-windows/properties/non_app_drains?
   value: "syslog-tls://doppler.service.cf.internal:6067"
+
+- type: replace
+  path: /addons/name=loggr-syslog-agent/jobs/name=loggr-syslog-agent/properties/drain_ca_cert?
+  value: "((log_cache_syslog_tls.ca))"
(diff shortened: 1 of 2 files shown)


davewalter:
Hi @kitemongerer & @heycait,

It looks like y'all included a version bump for `loggregator-agent-release` in this PR. In general, you don't need to do this as our [update-releases](https://release-integration.ci.cf-app.com/teams/main/pipelines/update-releases) pipeline automatically picks up new final releases and validates them (deploy + smoke-tests) before bumping the version on develop. For example, we [automatically consumed loggregator-agent-release v4.0.1 this morning](https://release-integration.ci.cf-app.com/teams/main/pipelines/update-releases/jobs/update-loggregator-agent/builds/35). The problem with including version bumps in PRs is that it doesn't cover all of the actions taken by the `update-releases` pipeline. Notably, we compile all releases in the base cf-deployment.yml manifest against the latest Linux stemcell and make them available in an [optional ops-file](https://github.com/cloudfoundry/cf-deployment/blob/master/operations/use-compiled-releases.yml), which is not something that you could include in the PR. The other problem is that it causes conflicts that prevent us from being able to merge a PR using Github. The only time that including a release version bump in a PR is desired/required is when our automation fails because a manifest/ops-file change is required.

Could you please remove the redundant version bump to fix the conflict in the base manifest so that this can be merged?

Thanks,
Dave

heycait:
Cool thanks for clarifying the process. We've removed the version changes to `cf-deployment.yml`

PR #830, opened 2019-10-10
Add Metric Store

toddboom:

This adds a single node Metric Store to the CF Deployment.

As metric support is phased out of Log Cache, users will need a new store for firehose metrics. Metric Store aims to fill this need by connecting directly to the RLP, durably storing all metrics for a foundation, and providing a PromQL-compatible query interface to access them.

https://www.pivotaltracker.com/story/show/167953063

We are just about to cut an updated release (v1.2.0) that should probably be specified in the manifest before this is merged. However, as I anticipate this PR will generate some discussion, we're getting it up here early so that we can start talking about it sooner.

- [ ] YES
- [x] NO

- [x] YES
- [ ] NO

This change increases VM footprint of cf-deployment.

Adds a Metric Store instance, which provides operators and users with a durable store of firehose metrics that can be queried via PromQL.

- [x] YES
- [ ] NO

This will add the `metric-store-release` BOSH release.

- [ ] GA'd feature/component

An admin user should be able to run the following curl command and see a sucessful response:

```
curl -H "Authorization: $(cf oauth-token)" -sS -G https://my.foundation.pivotal.io/api/v1/query --data-urlencode 'query=cpu`
```

/cc @colins @attack @hdub2 @josephgee @heidmotron


acosta11:
Unit tests failed due to known issue. Should now be resolved, so closing and re-opening to re-trigger tests.

toddboom:
@acosta11 thanks for retriggering - looks like it still failed. we'll investigate on this end.

weymanf:
Hey @toddboom Looks like unit tests are still failing for this PR. 
The changes you need to make are in this file: 
https://github.com/cloudfoundry/cf-deployment/blob/develop/operations/rename-network-and-deployment.yml

You have to add the instance to this list to change the network name. And we should be good to go.

operations/experimental/add-metric-store.yml
```diff
@@ -1,169 +0,0 @@
-- type: replace
```
weymanf: Looks like this operation of excluding the prom_scrapper from the metrics-store instance group was not replaced inside of the base manifest. Is this the correct behavior? Does the metrics-store now need to have the prom_scraper addon.

operations/experimental/add-metric-store.yml
```diff
@@ -1,169 +0,0 @@
...
-  path: /addons/name=prom_scraper/exclude?/instance_groups/-
-  value: metric-store
```
weymanf: Looks like this operation of excluding the prom_scrapper from the metrics-store instance group was not replaced inside of the base manifest. Is this the correct behavior? Does the metrics-store now need to have the prom_scraper addon?

julian-zucker:
Moved to https://github.com/cloudfoundry/cf-deployment/pull/843

PR #835, opened 2019-10-23
Add Ops Files to Disable v2 Logging Pipeline

chentom88:

Add ops files to disable the v2 logging pipeline jobs so users can prepare for it's eventual deprecation.

The v2 logging pipeline is going to be deprecated eventually. This is to prepare them for that deprecation.

https://www.pivotaltracker.com/story/show/169021301

- [ ] YES
- [x] NO

This breaks some functionality in the CLI. However, they have stories to convert to using log-cache instead.

- [x] NO

Introduce ops files to disable the v2 logging pipeline so teams can prepare for its eventual deprecation.
 
- [ ] YES - please specify
- [x] NO

- [ ] GA'd feature/component

If these ops files are deployed, the RLP should not be on the Doppler vm and the loggregator agent should not be on any VMs.

@MasslessParticle @mjseaman @pianohacker


operations/experimental/use-logcache-syslog-ingress.yml
```diff
@@ -21,7 +21,7 @@
     type: certificate
     options:
       ca: loggregator_ca
-      common_name: log-cache-syslog
```
weymanf: Is this a required changed for the removals to work? Or could this be a separate change.

kitemongerer: It can be a separate change but syslog ingress into log-cache needs to work in order to disable the v2 firehose and it won't without this change so they are semi-related. We can pull it into a different PR if you would prefer

weymanf: Thanks for the info. No I think this change is okay to be pulled in this PR since it's experimental.

weymanf:
Is there a plan for us to test this in our pipelines? It looks like this will eventually be GA'd and we would like CATs to pass before pulling this in. What are the CLI stories that need to be done to keep the CLI functionality?

chentom88:
@weymanf Would it be necessary to make it pass CATs being an experimental ops-file?

weymanf:
Ideally, experimental ops files shouldn't destroy existing functionality. I was just worried from this sentence in the PR:
"This breaks some functionality in the CLI. However, they have stories to convert to using log-cache instead."

If the purpose of this PRs is for CAPI/CLI team to deploy these opsfiles so they can test against it, then we will consider pulling it in. If all this does is shift a bunch of instances around and doesn't effect the `cf logs` command, then I'm also ready to pull this in.

kitemongerer:
@weymanf The purpose is for the CLI team and other users of the firehose to be able to use this ops file in their dev environments to validate that they don't depend on the firehose and will be unaffected by its deprecation. It will break the CLI in the short term but it will help them migrate away from the firehose. I'm also happy to jump on a Zoom if that will help clarify things

chentom88:
@weymanf Did you get the answers you needed from us? Is there anything else we should do to move this along?

weymanf:
I think our PM is going to discuss this with your team. @Syerram

chentom88:
Closing this PR and replacing with https://github.com/cloudfoundry/cf-deployment/pull/857

PR #843, opened 2019-10-31
Move metric store opsfile out of experimental

julian-zucker:

This adds a single node Metric Store to the CF Deployment.

As metric support is phased out of Log Cache, users will need a new store for firehose metrics. Metric Store aims to fill this need by connecting directly to the RLP, durably storing all metrics for a foundation, and providing a PromQL-compatible query interface to access them.

https://www.pivotaltracker.com/story/show/167953063

We are just about to cut an updated release (v1.2.0) that should probably be specified in the manifest before this is merged. However, as I anticipate this PR will generate some discussion, we're getting it up here early so that we can start talking about it sooner.

- [ ] YES
- [x] NO

- [x] YES
- [ ] NO

This change increases VM footprint of cf-deployment.

Adds a Metric Store instance, which provides operators and users with a durable store of firehose metrics that can be queried via PromQL.

- [x] YES
- [ ] NO

This will add the `metric-store-release` BOSH release.

- [ ] GA'd feature/component

An admin user should be able to run the following curl command and see a sucessful response:

```
curl -H "Authorization: $(cf oauth-token)" -sS -G https://my.foundation.pivotal.io/api/v1/query --data-urlencode 'query=cpu`
```

/cc @colins @attack @hdub2 @josephgee @heidmotron @toddboom


weymanf:
Hello. Should I close https://github.com/cloudfoundry/cf-deployment/pull/830? Since this looks like a duplicate?

toddboom:
@weymanf I believe @julian-zucker closed it already. We opened a new PR to reduce the clutter, since it seems like the preferred path is to move the ops file out of experimental after talking to @Syerram a couple weeks ago.

toddboom:
Also, we are just about to cut a new OSS release in the next few days, so we can probably wait to merge this until that is ready to go. It will allow us to clean up the ops file just a bit.

weymanf:
Is this a GA'd (validated in your pipeline) version of the metrics-store?

PR #848, opened 2019-11-19
Fix metric-store opsfile to be a usable opsfile

aminjam:

The 2 properties described here are not valid in the metric-store
release and the intended values are the default values in metric-store
job. This causes the yaml parse failure when using this opsfile.

egress_port -> port
health_addr -> health_port

- [x] YES
- [ ] NO

- [x] NO

> Fix metric-store opsfile

- [ ] YES - please specify
- [x] NO

- [x] GA'd feature/component

Using the opsfile in cf-deployment should succeed. 

- [] **Urgent** - unblocks current or future work

@cf-diego

### DIFF
--- a/operations/use-metric-store.yml
+++ b/operations/use-metric-store.yml
@@ -17,8 +17,6 @@
     jobs:
     - name: metric-store
       properties:
-        egress_port: 8080
-        health_addr: localhost:6060
         metric_store_server:
           tls:
             ca_cert: ((metric_store_server.ca))


aminjam:
We realized that our CI was stuck at an earlier commit (https://github.com/cloudfoundry/cf-deployment/commit/d9effe4f4a0741d73aa3602b140b21469ef7f2e7) where we saw the error. We think this PR is still valuable given the the two properties are not used, but it's not critical where it caused an error.

sunjayBhatia:
@pivotaljohn Commit amended and repushed

PR #855, opened 2019-12-12
Optimize deployment parallelization

jpalermo:

> Remove extra `serial: true` statements to speed parallel deployment.

> deploy fast enough...   A test deployment I performed went from ~27 minutes to ~21 minutes. Actual deployment times gains will vary depending on instance counts and other included ops files.

- [ ] YES
- [X] NO

The changes were successfully BOSH deployed. At that point all the components are working so there should be no reason for behavior to change.

- [X] NO

> Better optimize deployment parallelization to speed the deployment processes.

- [ ] YES - please specify
- [X] NO

- [ ] GA'd feature/component

> Behavior should remain the same. Deployment time should be reduced.

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -269,8 +269,6 @@ instance_groups:
   instances: 1
   vm_type: minimal
   stemcell: default
-  update:
-    serial: true
   networks:
   - name: default
   jobs:
@@ -431,8 +429,6 @@ instance_groups:
   instances: 2
   vm_type: small
   stemcell: default
-  update:
-    serial: true
   networks:
   - name: default
   jobs:
@@ -1089,8 +1085,6 @@ instance_groups:
   vm_extensions:
   - diego-ssh-proxy-network-properties
   stemcell: default
-  update:
-    serial: true
   networks:
   - name: default
   jobs:
@@ -1568,8 +1562,6 @@ instance_groups:
   instances: 2
   vm_type: minimal
   stemcell: default
-  update:
-    serial: true
   networks:
   - name: default
   jobs:
--- a/operations/experimental/fast-deploy-with-downtime-and-danger.yml
+++ b/operations/experimental/fast-deploy-with-downtime-and-danger.yml
@@ -8,23 +8,11 @@
   value: 100%
 
 - type: replace
-  path: /instance_groups/name=smoke-tests/update/serial
-  value: false
-- type: replace
   path: /instance_groups/name=database/update/serial
   value: true
-- type: replace
-  path: /instance_groups/name=diego-api/update/serial
-  value: false
 - type: replace
   path: /instance_groups/name=singleton-blobstore/update/serial
   value: false
 - type: replace
   path: /instance_groups/name=router/update/serial
-  value: false
-- type: replace
-  path: /instance_groups/name=scheduler/update/serial
-  value: false
-- type: replace
-  path: /instance_groups/name=log-api/update/serial
   value: false


cf-deployment.yml
```diff
@@ -269,8 +269,6 @@ instance_groups:
   instances: 1
   vm_type: minimal
   stemcell: default
-  update:
-    serial: true
```
jpalermo: `serial: true` does not make sense for `errand` instance groups I believe.

cf-deployment.yml
```diff
@@ -431,8 +429,6 @@ instance_groups:
   instances: 2
   vm_type: small
   stemcell: default
-  update:
-    serial: true
```
jpalermo: This `serial: true` only really ensures that the `diego-api` instance group happens before the `uaa` instance group. I performed a deploy where I move the position of `diego-api` after `uaa`, but left `serial: true` enabled. The deploy was successful, showing that `uaa` has no dependency on `diego-api` and it should be safe to deploy them in parallel.

cf-deployment.yml
```diff
@@ -1089,8 +1085,6 @@ instance_groups:
   vm_extensions:
   - diego-ssh-proxy-network-properties
   stemcell: default
-  update:
-    serial: true
```
jpalermo: This `serial: true` here seems to ensures that the `scheduler` instance group happens after the `api` and `cc-worker` instance groups. This might have been required at some point in time due to jobs needing the `ccdb` to be in place, or possibly the `cloud_controller_ng` job to be running before one of the `scheduler` jobs was started.

I performed a deploy where I move the position of `scheduler` before the `api` instance group, but left `serial: true` enabled. The deploy was successful, showing that `scheduler` has no dependency on `api` or `cc-worker` instance groups and it should be safe to deploy them in parallel.

cf-deployment.yml
```diff
@@ -1568,8 +1562,6 @@ instance_groups:
   instances: 2
   vm_type: minimal
   stemcell: default
-  update:
-    serial: true
```
jpalermo: My best guess about the `serial: true` here is that it ensures that the `log-api` instance group happens after the `doppler` instance group.

I performed a deploy from scratch where I move the position of `log-api` instance group before the `doppler` instance group, but left `serial: true` enabled. The deploy was successful, showing that `log-api` instance group has no dependency on the `doppler` instance groups and it should be safe to deploy them in parallel.

davewalter:
Hi @jpalermo,

We did some validation of this change in an upgrade scenario with uptimer to see what effect it has on the platform while components are rolling. While we didn't see any application availability issues, we did see an increase in app pushability failures, as well as issues related to app log retrieval/streaming. At this point, we are concerned about the stability of cf-deployment for our long-time users and would want to validate these changes over a period of time by introducing them in an ops-file that we include in our CI.

We're curious as to why you aren't just using the `fast-deploy-with-downtime-and-danger` ops-file to speed up deployment? Are you trying to cover the time it takes to upgrade a deployment as well as fresh installations?

Regards,
Dave and Sebastian (@vitreuz)

jpalermo:
That's interesting, do you have any data on which components were being updated when the instability occurred? I'd be interested to dig into what was causing that downtime.

`fast-deploy-with-downtime-and-danger` will cause downtime if I'm reading it correctly. I'd like to speed up the deployment time by only serializing components that actually have dependencies on other components, without sacrificing uptime.

Once upon a time cf deploys were done completely serially and it took a VERY long time. We're in a better spot now, but I think there are still some optimizations that can be made, which do save a non-trivial amount of time with each deploy.

jamespollard8:
@jpalermo apologies for the response time here. It's been hard for us to prioritize cf-d work. In the next week we should be able to kick-off testing with this again and get new data.

davewalter:
Thanks @jpalermo. Our additional testing didn't show any additional downtime during upgrade so we are going to take the opportunity to get this into v13.

PR #868, opened 2020-02-21
Add the disabled syslog agent ops file

Benjamintf1:

Add an ops file that adds agents to all vms for the purpose of egressing application logs to syslog, but do not enable their use. Can be used before add-syslog-agent.yml to decrease downtime.

Allows use before add-syslog-agent.yml to decrease downtime for drains.

- [x] YES
- [ ] NO

- [x] NO

Adds an ops file to allow reduced downtime for syslog drains. 

- [ ] YES - please specify
- [x] NO

- [ ] GA'd feature/component

Create a cf-d normally. Create a cf-syslog drain. Use the add-disabled-syslog-agent.yml ops file, then use add-syslog-agent.yml in new seperate deploy. See low downtime on syslog endpoint. 

@pianohacker @chentom88 @Benjamintf1


davewalter:
Hi @Benjamintf1,

It looks like it is a duplicate of the `add-syslog-agent.yml` ops-file, with the addition of `enabled: false` to the `loggr-syslog-binding-cache` instance, which seems less than ideal as they could easily get out of sync as future changes are made. I would suggest a more simple `disable-syslog-agent.yml` ops-file that looks something like this:
```
- type: replace
  path: /instance_groups/name=scheduler/jobs/name=loggr-syslog-binding-cache/properties/enabled
  value: false
```

This ops-file could then be interpolated after `add-syslog-agent.yml` in the first deploy, and removed from the subsequent deploy to enable the agent. Let me know what you think.

Regards,
Dave

Benjamintf1:
@davewalter Not exactly. It *also* doesn't remove the syslog adapters/release/etc. It's less then ideal, but
1) We are planning on removing *both* of these ops files in v13
2) it prevents breaking people who are currently using add-syslog-agent for a decomposed opsfile
3) it means that we don't have to *add back* everything that gets removed in the subsequent opsfile if we wanted to build on top of add-syslog-agent.yml(namely, the adaptors).

davewalter:
We updated the wording a little bit to make it more clear the use case for this ops-file. Thanks for the PR.

If you are planning to PR removing both for cf-deployment v13, we would recommend making the PR if you haven't already as we are starting to plan for that major release.

Cheers
Dave and @angelachin

PR #872, opened 2020-03-11
Remove unused variable cf_bosh_password

andrew-edgar:
### WHAT is this change about?

found that there is an unused variable called "cf_bosh_password".  This is just removing that from the manifests.

### What customer problem is being addressed? Use customer persona to define the problem e.g. Alana is unable to...

Nothing really just cleaning up the manifest so there are no unused variables.

### Has a cf-deployment including this change passed [cf-acceptance-tests](https://github.com/cloudfoundry/cf-acceptance-tests)?

- [x] YES
- [ ] NO

### Does this PR introduce a breaking change? Please take a moment to read through the examples before answering the question.

- [ ] YES - please choose the category from below. Feel free to provide additional details.
- [x] NO

### How should this change be described in cf-deployment release notes?

removed unused variable cf_bosh_password

### Does this PR introduce a new BOSH release into the base cf-deployment.yml manifest or any ops-files?

- [ ] YES - please specify
- [x] NO

### Does this PR make a change to an experimental or GA'd feature/component?

- [ ] experimental feature/component
- [ ] GA'd feature/component

### Please provide Acceptance Criteria for this change?

`bosh variables | grep cf_bosh_password`  will return no variables.

### What is the level of urgency for publishing this change?

- [ ] **Urgent** - unblocks current or future work
- [x] **Slightly Less than Urgent**

PR #875, opened 2020-03-30
Revert "Remove unused variable cf_bosh_password"

Syerram:
Reverts cloudfoundry/cf-deployment#872

The change is a breaking change to the main manifest. We can put this in v13 backlog.


andrew-edgar:
so my PR was merged but this reverted it.  so what is the status of my PR to remove this getting in cf-deployment?  why is my PR closed when it was never integrated 
https://github.com/cloudfoundry/cf-deployment/pull/872

PR #877, opened 2020-04-15
Enhancement: NATS servers are now addressed with hostnames instead of IP addresses

KauzClay:
Instead of using IPs as addresses for nats servers, use hostnames instead. This allows nats to take advantage of bosh dns, so that when instances roll, IP changes won't cause breaks in communication. This also lets us avoid nats split-brain problems by taking advantage of bosh dns to prefix hostnames with instance ids, so that each nats server can have a unique dns resolution.

Alana is able to roll NATS vms with minimal interruption to control plane communication.

This is more of an enhancement, though we did have a related inquiry recently that we happen to fix with this, we believe: https://cloudfoundry.slack.com/archives/CFX13JK7B/p1586368934041100

Mainly, this was done in response to a recent [failure](https://release-integration.ci.cf-app.com/teams/main/pipelines/update-releases/jobs/update-nats/builds/20) in RelInt pipelines for nats v34.

The story for this work is here: [#172315755](https://www.pivotaltracker.com/story/show/172315755)

More information on the split-brain solution that is the main feature for v34 can be found here: [#172137909](https://www.pivotaltracker.com/story/show/172137909)

- [ ] YES
- [ ] NO

- [X] NO

NATS servers are now addressed with hostnames instead of IP addresses

- [ ] YES - please specify
- [X] NO

- [X] GA'd feature/component

Given I have a cf-deployment with one or multiple NATs instances
When I ssh onto the NATS vm
And I `cat /var/vcap/jobs/nats/config/nats.conf`
Then I see that the `routes` section of the config contains addresses with `<instance-id>.nats.service.cf.internal` as the hostname.

@kauana @ameowlia 
#networking in cf slack, use the @tiger-eyes alias.

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -260,6 +260,21 @@ addons:
           deployment: cf
           network: default
           domain: bosh
+      - domain: nats.service.cf.internal
+        targets:
+        - query: '*'
+          instance_group: nats
+          deployment: cf
+          network: default
+          domain: bosh
+      - domain: _.nats.service.cf.internal
+        targets:
+        - query: '_'
+          instance_group: nats
+          deployment: cf
+          network: default
+          domain: bosh
+
 
 instance_groups:
 - name: smoke-tests
@@ -308,6 +323,7 @@ instance_groups:
       nats:
         password: "((nats_password))"
         user: nats
+        hostname: nats.service.cf.internal
 - name: adapter
   azs:
   - z1


jamespollard8:
It looks like this may have broken some of our deploy jobs in CI. For example [fresh-deploy](https://release-integration.ci.cf-app.com/teams/main/pipelines/cf-deployment/jobs/fresh-deploy/builds/2701#L5e2dcdc6:4734)

Our team will be taking a look this afternoon.

davewalter:
We needed to add the new DNS alias to the `rename-network-and-deployment.yml` ops-file.

KauzClay:
Thanks @jamespollard8 @davewalter for taking care of that!

So, how can I check the progress of our nats release v34? I've been watching https://release-integration.ci.cf-app.com/teams/main/pipelines/update-releases/jobs/update-nats/builds/22

But I haven't seen a run with our commit in release-candidate yet

jamespollard8:
We're working on this some more this morning. I've added a task to our story to remember to circle back here with an update for you.

IIUC, teammates found that the issue was that this change affected an ops file that requires a small change.

We're also seeing [weirdness with route availability](https://release-integration.ci.cf-app.com/teams/main/pipelines/cf-deployment/jobs/bbr-deploy/builds/773) but only on CI, it's working fine on our local machines. We suspect it's something about our environment.

KauzClay:
@jamespollard8 we noticed that the PR you made to the opsfile only included the entry for "*" querys for the bosh dns-alias. But in our PR, we had two entries https://github.com/cloudfoundry/cf-deployment/pull/877/files

I think we need the "_" entry also, because we are trying to leverage using instance ids on the hostnames. The usages for "_" says it is for this case (https://bosh.io/docs/dns/).

Maybe that is contributing to the weirdness you're seeing?

cc @ameowlia

PR #899, opened 2020-08-03
Fix cc_deployment_updater always refers to only one mysql instance even if ccdb is scaled out

kei-yamazaki:

Even if you scale out CCDB with `operations/scale-database-cluster.yml`, cc_deployment_updater will always reference only CCDB with `index: 0`.
Therefore, cc_deployment_updater becomes down just by the appearance of `index: 0` in Percona XtraDB Cluster is down (Percona XtraDB Cluster is in a healthy state at this time).

This is due to the code below.
https://github.com/cloudfoundry/capi-release/blob/4899333c053273e8a24036c1f803adb70c67218f/jobs/cc_deployment_updater/templates/cloud_controller_ng.yml.erb#L49-L65

Changed cc_deployment_updater to refer to CCDB by ip address to use BOSH DNS record ( `sql-db.service.cf.internal` ).

Alana is now able to use cf even if `index: 0` of CCDB cluster goes down.

N/A

- [x] YES
- [ ] NO

- [x] NO

Fix cc_deployment_updater always refers to only one mysql instance even if ccdb is scaled out

- [ ] YES - please specify
- [x] NO

- [x] GA'd feature/component

After building CF with `operations/scale-database-cluster.yml`, `bosh stop database/0` is possible  `cc_deployment_updater` job in scheduler vm must be started.

I'm an individual :)

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -1232,16 +1232,7 @@ instance_groups:
           ca_cert: "((cc_tls.ca))"
           private_key: "((cc_tls.private_key))"
           public_cert: "((cc_tls.certificate))"
-      ccdb:
-        databases:
-        - name: cloud_controller
-          tag: cc
-        db_scheme: mysql
-        port: 3306
-        roles:
-        - name: cloud_controller
-          password: ((cc_database_password))
-          tag: admin
+      ccdb: *ccdb
   - name: service-discovery-controller
     properties:
       dnshttps:


davewalter:
Hi @kei-yamazaki,

As an FYI that we thought you would be curious to hear about, we just wanted to let you know about issues we ran into in our CI once this had been merged. We found that there were other ops-files related to non-default database configurations that required minor attention. See 71b99097 for details. It is worth noting that we only caught this because of our CI testing, and would probably have missed it ourselves when making the change.

Thanks again for the change.
Dave and @jamespollard8

PR #903, opened 2020-09-10
Add SAN to cc_tls certificate

reid47:

We need to add a SAN to the `cc_tls` cert. This is so we can safely upgrade to Go 1.15 in `capi-release`. Without this, the cc_uploader will complain in Go 1.15 that CC's cert relies on the deprecated pattern of specifying a common name without a SAN.

> _Understanding why this change is being made is fantastically helpful. Please do tell..._

This change allows `capi-release` developers to continue to auto-bump their Go dependency to 1.15+ (currently it is pinned at 1.14.x). If `capi-release` bumps Go without this change, the cc_uploader will fail to talk to CC, and `cf push`es will fail at the staging step (affecting Codys).

> _Include any links to other PRs, stories, slack discussions, etc... that will help establish context._

- Golang release notes describing this change: https://golang.org/doc/go1.15#commonname
- Slack thread about this PR: https://cloudfoundry.slack.com/archives/C2U7KA7M4/p1599754038020000

- [x] YES
- [ ] NO

- [ ] NO -

Please see [this Slack thread](https://cloudfoundry.slack.com/archives/C2U7KA7M4/p1599754038020000). This PR on its own is not a breaking change. *However*, the bump to go 1.15 which this is in preparation for will cause a breaking change to `cf push`. We would like to use this PR to open a conversation about potential next steps:

a) Generate certificate property as-is and ask operators to delete the variable `cc_tls` prior to deploying so that the cert will be re-generated.
b) Add `update_mode: converge` to the `cc_tls` property to ensure the cert will be re-generated. (Potentially remove it in a future release).

Regardless of which mitigation we introduce, we may want to save the GoLang 1.15 upgrade to the next minor version.

>    - `./addons`
>
> _If you're promoting an experimental Ops-file (or removing one), Please follow the [Ops-file workflows](https://github.com/cloudfoundry/cf-deployment/blob/master/ops-file-promotion-workflow.md)._

> Add Subject Alternative Name (SAN) to `cc_tls` certificate in preparation for update to GoLang 1.15.

- [ ] YES - please specify
- [x] NO

- [x] GA'd feature/component

This is the simplest varification but you can also provide additional commands to test that the variable holds the desired value.
* `bosh ssh api`
* `sudo su`
* `openssl s_client -connect cloud-controller-ng.service.cf.internal:9023  | openssl x509 -noout -text`
* View the output. Make sure that the following property is present:
```
X509v3 Subject Alternative Name: 
  DNS:cloud-controller-ng.service.cf.internal
```

@cloudfoundry/v3-acceleration-team-vat

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -2201,6 +2201,7 @@ variables:
     common_name: uaa_login_saml
 - name: cc_tls
   type: certificate
+  update_mode: converge
   options:
     ca: service_cf_internal_ca
     common_name: cloud-controller-ng.service.cf.internal
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -2204,6 +2204,8 @@ variables:
   options:
     ca: service_cf_internal_ca
     common_name: cloud-controller-ng.service.cf.internal
+    alternative_names:
+    - cloud-controller-ng.service.cf.internal
     extended_key_usage:
     - client_auth
     - server_auth


sunjayBhatia:
We are running into this issue with diego-release bumping to go 1.15 and would expect loggregator components will do as well since loggregator certificate variables have only common names set in the cf-d variable configuration

sunjayBhatia:
cc @cloudfoundry/cf-diego

jamespollard8:
Thanks @reid47 - we appreciate your thoroughness with this!

I'm also adding a similar slack thread for reference: https://cloudfoundry.slack.com/archives/C2U7KA7M4/p1600285696032100?thread_ts=1600285368.029400&cid=C2U7KA7M4

PR #926, opened 2021-05-28
move aggregate drains to the loggr-syslog-binding-cache job

Benjamintf1:
- this should speed up deploys considerably when the only change is a aggregate drain
- this should also make aggregate drains spread over deployments without a additional deploy

move aggregate drains to the loggr-syslog-binding-cache job
- this should speed up deploys considerably when the only change is a aggregate drain
- this should also make aggregate drains spread over deployments without a additional deploy if those other deployments also make this change

- aggregate drains require deploying every vm on the system

- [x] YES
- [ ] NO

- [x] NO

- operators can now change aggregate drains(if specified in the loggr-syslog-binding-cache job), on one vm rather then changing every vm

- [ ] YES - please specify
- [x] NO

- [x] GA'd feature/component

- cats succeeds when the opsfile to use syslog for log-cache is used

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -2812,9 +2812,9 @@ releases:
   version: 75.1.0
   sha1: d23e29e8e16c127e640ea992aa5e7ac4184000c7
 - name: loggregator-agent
-  url: https://bosh.io/d/github.com/cloudfoundry/loggregator-agent-release?v=6.2.0
-  version: 6.2.0
-  sha1: 7210bac9c456bf20fd6de2175c562e443f249249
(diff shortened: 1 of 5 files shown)


davewalter:
Hi Ben,

Thanks for the contribution. This looks good overall. One point worth noting is that we generally don't accept version bumps in PRs. The reason for this is that we maintain compiled releases in a separate ops-file, which is impossible to update in the PR (you can't put the new compiled release in the S3 bucket). However, in this case, it doesn't matter as we have already consumed the version bump via our [CI pipeline](https://release-integration.ci.cf-app.com/teams/main/pipelines/update-releases/jobs/update-loggregator-agent/builds/65).

Dave

PR #949, opened 2022-02-01
split log-cache from doppler, use syslog ingress

mkocher:
Making this change for a few reasons:
- The scaling needs of dopplers and log-cache are often different, so
  grouping them together can be problematic. Dopplers are limited to ~40
  instances and some high traffic foundations need larger log-cache
  instance groups.
- Syslog ingress eliminates the load on dopplers and traffic controllers
  to get envelopes to log-cache. This increases the load slightly on
  diego cells, and eliminates significant load on dopplers/tc's.

It's recommended after deploying this change to evaluate the memory
allocated doppler nodes and switch them to compute heavy instances and
deploy log-cache to high memory instances.

- [X] YES
- [ ] NO

- [ ] NO

>

Log Cache is now deployed separately from Doppler on its own instance group. Operators should consider scaling the memory on Doppler nodes down and using high memory Log Cache nodes. Operators should notice reduced CPU usage on Doppler & Traffic Controller and slight increase CPU usage by Syslog Forwarder.

- [ ] YES - please specify
- [X] NO

- [ ] GA'd feature/component

`bosh vms` should show a log-cache instance group.
`bosh ssh log-cache` should be a vm running log-cache and assorted processes
`bosh ssh doppler` should show a vm not running log cache when executing `sudo monit summary`

@rroberts2222


operations/experimental/use-logcache-syslog-ingress.yml
```diff
@@ -1,54 +1,4 @@
+# Has been integrated into cf-deployment.yml
+#
+# Please delete this file in the future
```
davewalter, line 3: Since this is already a breaking change, you can delete this ops-file (and the corresponding entries in the `README.md` and `units/tests/experimental_test/operations.yml` files). We'll just note the deletion in the release notes.

operations/experimental/use-logcache-syslog-ingress-windows2019.yml
```diff
@@ -1,4 +1,4 @@
+# Has been integrated into cf-deployment.yml
+#
+# Please delete this file in the future
```
davewalter, line 3: Since this is already a breaking change, you can delete this ops-file (and the corresponding entries in the `README.md` and `units/tests/experimental_test/operations.yml` files). We'll just note the deletion in the release notes.

davewalter (changes requested):
Thanks @mkocher & @rroberts2222. This looks good overall. I just had one change to request regarding the deprecated experimental ops-files.

operations/experimental/use-logcache-syslog-ingress-windows2019.yml
mkocher: 👍

operations/experimental/use-logcache-syslog-ingress.yml
mkocher: 👍

davewalter (approved):
LGTM

mkocher:
Noticed that cc defaults to using doppler as the stats server which is breaking stats. We'll push a fix.

cf-deployment.yml
```diff
@@ -338,7 +339,13 @@ addons:
           deployment: cf
           network: default
           domain: bosh
-
+      - domain: log-cache.service.cf.internal
+        targets:
+        - deployment: cf
```
ctlong, line 344: Key-ordering is inconsistent with the other bosh-dns entries.

(I admire the commitment to alphabetical though)

cf-deployment.yml
```diff
@@ -949,6 +949,8 @@ instance_groups:
         staging_upload_user: staging_user
         staging_upload_password: "((cc_staging_upload_password))"
         temporary_use_logcache: true
+        logcache:
+          host: log-cache.service.cf.internal
```
Gerg, line 953: Is `log-cache.service.cf.internal` available on older versions?

Wondering about what happens during an upgrade if CC starts talking to `log-cache.service.cf.internal` and it isn't available yet OR if CC is still trying to reach log cache on `doppler.service.cf.internal:8080` and it is no longer there. 

If so, does it just mean app metrics aren't available during the upgrade? Are there other effects?

cf-deployment.yml
```diff
@@ -1426,27 +1436,16 @@ instance_groups:
...
         key: "((loggr_udp_forwarder_tls.private_key))"
         server_name: loggr_udp_forwarder_metrics
-- name: doppler
+- name: log-cache
   azs:
   - z1
   - z2
-  instances: 4
-  vm_type: minimal
+  instances: 1
```
ctlong, line 1443: Why 1 instance instead of more?

ctlong: Doppler is set to 4.

cf-deployment.yml
```diff
@@ -1473,21 +1472,17 @@ instance_groups:
...
+        cert: "((log_cache_syslog_tls.certificate))"
+        key: "((log_cache_syslog_tls.private_key))"
       metrics:
-        ca_cert: ((log_cache_nozzle_metrics_tls.ca))
-        cert: ((log_cache_nozzle_metrics_tls.certificate))
-        key: ((log_cache_nozzle_metrics_tls.private_key))
-        server_name: log_cache_nozzle_metrics
-      logs_provider:
-        tls:
-          ca_cert: ((logs_provider.ca))
```
ctlong, line 1487: It appears that we can remove the logs_provider credhub variable. Can't seem to find any other reference to it.

cf-deployment.yml
```diff
@@ -1526,6 +1521,27 @@ instance_groups:
         client_secret: ((uaa_clients_doppler_secret))
         internal_addr: https://uaa.service.cf.internal:8443
     release: log-cache
+- name: doppler
+  azs:
+  - z1
+  - z2
+  instances: 4
```
ctlong, line 1528: Could we reduce the number of doppler instances to match the gain in log-cache instances (to some extent)? Is that a bad idea?

cf-deployment.yml
ctlong: If we do change this instance number, we should probably add a line to the bosh-lite ops file to put the instance count back to 1.

ctlong (changes requested):
Some other operators may not want to move to the syslog ingress model, and to accommodate them it is probably a good idea to include some ops files that would restore the previous log-cache-nozzle / RLP ingress.

cf-deployment.yml
rroberts2222: We couldn't find a reason to have more than one (log cache does not replicate data, so having two would just add more memory, not reliability) and wanted to keep a minimal footprint.

cf-deployment.yml
rroberts2222: Yes, let's reduce to 3 so log-cache vms + doppler vms still equals 4.

ctlong (approved):

cf-deployment.yml
ctlong: @Gerg does that consequence seem okay to you? Is there anything else we can clear up?

ctlong (approved):

cf-deployment.yml
Gerg: Yea, that sounds acceptable. I'd definitely include it in the release notes, so people aren't surprised.

ctlong:
This PR can cause significant log downtime if operators are not prepared for it when they first upgrade to a release its shipped in (v18.0.0+).

There has been a lot of [conversation in slack](https://cloudfoundry.slack.com/archives/C2U7KA7M4/p1647455071443379?thread_ts=1647447521.012169&cid=C2U7KA7M4) about this so I wanted to repost some here for posterity. In order to minimize log downtime operators should deploy twice, once to make the new log cache service alias available to every VM, and again to actually deploy the change. The deployment order should be something along the lines of:
1. New Log Cache VMs
1. Scheduler VM: lets syslog agents in Diego Cells know to send logs/metrics to the new Log Cache (logs/metrics will still be sent to old log cache as long as syslog ingress to log cache was not already enabled)
1. API VMs: tells CAPI to use the new Log Cache rather than the old one
1. Doppler VMs: gets rid of the old log cache

To be safe, it's recommended to initially scale your new Log Cache up to the same instance count of Doppler VMs that you previously had, and then to look at metrics after the deploy in order to scale down both your Log Cache and your Doppler footprint.

acrmp:
Note that with this change the syslog agents running on the diego cells and other VMs need to be able to talk to the Log Cache Syslog Server on port 6067. Operators that are running diego cells within isolation segments may have to adjust their firewall rules.

PR #959, opened 2022-03-03
Remove unused q-s3.log-cache.default.cf.bosh

mkocher:
We cargo culted this from the ops file when we inlined it into
cf-deployment. It seems unnecessary since the default bosh dns strategy
is "smart" and clients should use the service alias.

The only consumers we expect this to affect would be someone who picked
this up in the last two weeks and started using the q-s3 domain instead
of the service alias, which we assume is a null set.

Removing alias that is not currently renamed by the "rename network" ops file so is kind of broken now.

Unable to rename the network using cf-deployment. 

https://cloudfoundry.slack.com/archives/C2U7KA7M4/p1646315931755709?thread_ts=1645812408.133929&cid=C2U7KA7M4

- [ ] YES
- [X] NO

- [X] YESish
- [ ] NO

> 4. modifies the name or deletes a property of a ~~job or instance group~~ certificate in the main manifest

Removes q-s3.log-cache.default.cf.bosh alias which we believe is unused.

- [ ] YES - please specify
- [X] NO

- [X] GA'd feature/component

> _Few examples_

@andrew-edgar  @Benjamintf1 

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -2266,7 +2266,6 @@ variables:
     ca: loggregator_ca
     common_name: log-cache.service.cf.internal
     alternative_names:
-      - "q-s3.log-cache.default.cf.bosh"
       - "log-cache.service.cf.internal"
     extended_key_usage:
       - server_auth


ctlong:
What about the possibility that a separate bosh deployment wants to communicate directly with Log Cache? The service aliases are currently only setup as add-ons in the cf deployment, not in the bosh runtime-config, so they won't be passed to other deployments.

Feel free to tell me this doesn't apply. I'm not entirely sure I'm referring to a valid use case 😅

Benjamintf1:
Largely speaking, I think any environment set up that way is likely and liable to be broken in this state anyways. It's a good question though. I'm not really sure if anything works at all without duplicating the bosh-dns-alias's anyways, given how widely they're used. 

Which is to say, It's very much possible, but I don't think we're necissarily in the position to "just fix it" well in any way anyways.

ctlong (approved):

davewalter (approved):

PR #962, opened 2022-03-09
Add database_encryption to cc_deployment_updater

cagiti:

This change adds the `database_encryption` YAML block to the `cc_deployment_updater` job. This might not be required, but we're assuming that if you've configured encryption key rotation you should specify the `database_encryption` block. The spec in the `cc_deployment_updater` seems to suggest this can be configured:

https://github.com/cloudfoundry/capi-release/blob/develop/jobs/cc_deployment_updater/spec#L105-L110

And since this is configured for all other CC properties it seems sensible to maintain consistency.

We've added it to reduce any confusion surrounding the cc properties configuration within the deployment manifest. By aligned all sections which specify `cc.properties` it reduces any confusion when rotating encryption keys for the cloud controller database.

- [x] NO

This change aligns the CC properties in terms of the Encryption key configuration for each CC instance group(api, cc-worker, scheduler).

- [ ] YES - please specify
- [x] NO

- [ ] GA'd feature/component

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -1256,6 +1256,7 @@ instance_groups:
     properties:
       cc:
         db_encryption_key: ((cc_db_encryption_key))
+        database_encryption: *cc-database-encryption
         mutual_tls:
           ca_cert: "((cc_tls.ca))"
           private_key: "((cc_tls.private_key))"


davewalter (approved):
Looks good.

I did notice that the `cloud_controller_internal` [link](https://github.com/cloudfoundry/capi-release/blob/e03d9c9723bac52dff4396f185913af55cc8c0ce/jobs/cloud_controller_ng/spec#L143-L145) provided by the `cloud_controller_ng` job does include both the `database_encryption` and `db_encryption_keys` specs. However, none of the CAPI jobs make use of either of those properties from the link.

PR #981, opened 2022-07-11
New ops-file addon to deploy minimum size to pass GCP CATS with 12 nodes

dsboulder:

The new standardized/efficient BBL VM sizes for CF-D are definitely not enough to pass CATS with any parallelism in CF-D Single AZ Scale-To-One. The diego cell capacity is now too small, a single VM with only 2 CPUs and 16GB. 

The solution suggested here is an additional ops file which can be _layered_ onto the singleAZ ops-file in order to deploy a somewhat large deployment. The larger deployment uses 2 API servers instead of 1, and 2 diego cells which have been resized to have more cores, now with 16GB and *4 CPUs*. 

_Developers of CF_ would like to ability to deploy a cost-effect CF-D by default using the new BBL sizes. The default CF-D configuration is barely enough to pass CATS with 12 nodes, and the single AZ deployment option is no longer sufficient. We'd like to give CF-D operators/CF developer the ability to use new efficient VM sizes by default that can pass CATS, but also have these 2 additional deployment options:
1. A single-AZ configuration that's as minimal as possible, likely for sandboxes and general playing with
2. A single-AZ configuration that's still fairly minimal, but will consistently pass CATS with 12 nodes on GCP.

https://cloudfoundry.slack.com/archives/C033ALST37V/p1657201388772499

- [x] YES
- [ ] NO

- [x] NO

>    - `./addons`
>

> Alana can choose an additional ops-file "addon" to scale CF-D to a "Single AZ" configuration with enough compute power to pass parallel CATS on GCP.

- [ ] YES - please specify
- [x] NO

- [ ] GA'd feature/component

> _Few examples_

@mkocher helped validate that this 2 x API server + 2 x diego-cell on e2-standard-4 does a good job passing CATS.


operations/scale-to-one-az-addon-parallel-cats.yml
```diff
@@ -0,0 +1,13 @@
+---
+# Use this after the scale-to-one-az ops file to scale up to minimal size
+# in order to run CATS quickly without flakes and 12 threads
+- type: replace
```
ctlong, line 4: @dsboulder what do you think about placing the CATs ops file in the `test` directory with the other ops files that are for testing, rather than production, purposes?

operations/scale-to-one-az-addon-parallel-cats.yml
```diff
@@ -0,0 +1,13 @@
...
+  path: /instance_groups/name=diego-cell/instances
+  value: 2
+- type: replace
```
ctlong, line 7: This file needs a unit test.

operations/scale-to-one-az-addon-parallel-cats.yml
dsboulder: Yep, great idea. Done!

PR #991, opened 2022-09-21
Drop unused application_ca

ae-govau:
This is only used to signed the diego_instance_ca, yet is never referenced as a trusted CA.

Currently this is the only instance of a CA that signs another CA, and this complicates an already complicated cerfificate rotation process.

By removing this CA signing a CA, the certificate rotation process is simplified.

See related issue https://github.com/cloudfoundry/credhub/issues/377

This removes `application_ca`, which is not referenced directly in any configuration.

The typical 4 step certificate rotation process outlined at <https://github.com/pivotal/credhub-release/blob/main/docs/ca-rotation.md> does not appear to work in a fully reliable fashion if CA 1 signs CA 2 which signs leafs. Sometimes it may work, depending on the order the transitional flags are added, but not always.

It is possible to follow those steps and be left in a situation where CA 2 is signed up with an out of date version of CA 1, leading to unexpected certificate expiration of a leaf due to an expired CA.

This is the only instance on `cf-deployment` of a CA that signs another CA.

See: https://github.com/cloudfoundry/credhub/issues/377

- [ ] YES
- [X] NO

I'm assuming opening a PR will run the acceptance tests.

- [X] NO

Hopefully not.

This should be an internal detail. Sadly currently the abstraction can leak after a certificate rotation.

- [ ] YES - please specify
- [X] NO

- [X] GA'd feature/component

Existing deployments can be re-deployed without intervention.


ae-govau:
I've tested by applying a equivalent operator file to an existing install:

Operator file:

```yaml
- type: remove
  path: /variables/name=application_ca
- type: remove
  path: /variables/name=diego_instance_identity_ca/options/ca
```

Result (showing that these are removed from the manifest, but doesn't change any actual instances):

```
  variables:
  - name: diego_instance_identity_ca
    options:
-     ca: application_ca
- - name: application_ca
-   options:
-     common_name: appRootCA
-     is_ca: true
-   type: certificate
Task xxx

Task xxx | 06:20:25 | Preparing deployment: Preparing deployment (00:01:00)
Task xxx | 06:21:25 | Preparing deployment: Rendering templates (00:01:01)
Task xxx | 06:22:28 | Preparing package compilation: Finding packages to compile (00:00:01)

Task xxx Started  Wed Sep 21 06:20:25 UTC 2022
Task xxx Finished Wed Sep 21 06:22:29 UTC 2022
Task xxx Duration 00:02:04
Task xxx done
```

ae-govau:
Adding an `update_mode: converge` causes the expected small amount of carnage as it rolls through.

If one doesn't add that, do we know if `credhub` will get given the new signing CA when BOSH runs? (so that next regeneration will be from the new signing CA (self) rather than the old one (`application_ca`)?

philippthun (changes requested):
I don't think that this will work.

We are successfully rotating all CAs/certificates for years now without any downtimes. I don't remember all the details, but I'm pretty sure that removing the `application_ca` will not work. Did you test this, i.e. deploy a foundation without this CA and run a full rotation and various downtime checks?

philippthun:
And during the rotation of the application_ca (i.e. instance_identity_ca is replaced as written above), it must be ensured that app containers can still talk to each other (and verify their identity) even across rotation steps.

Example: When the instance_identity_ca is renewed (step 2, called intermediateCA in the following picture), this is rolled out sequentially to the Diego cells, meaning that there will be - at the same time - cells issuing app certs with the new instance_identity_ca and other cells still using the old instance_identity_ca. And this three-level-hierarchy ensures that app containers still can verify each others identity.
Validity Across Rotation Steps

ae-govau:
Thanks for your comments @philippthun . You're right - as-is this won't work.

I had assumed that the behaviour whereby `credhub` concatenates CAs would behave equally whether the CA was a different CA, or it was self-signed. It turns out it won't in the self-signed case.

So yes, we should close this out and not merge it.

While we're here - I will describe the precise issue we're seeing after our rotation process (where we think we are following: <https://github.com/pivotal/credhub-release/blob/main/docs/ca-rotation.md>):

```bash
# Let's generate 2 CAs, a root (self-signed) and an intermediate.
# Normally we'd do with BOSH variables, but we do here manually to demonstrate
credhub generate --type certificate --name /test/root --common-name /test/root --is-ca --self-sign # cQc
credhub generate --type certificate --name /test/intermediate --common-name /test/intermediate --is-ca --ca /test/root # WZU

# Now, let's rotate certs.

cert_names="/test/root /test/intermediate"

# Step 1: Set transitional flag on each
for cert_name in ${cert_names} ; do
     cert_id="$(credhub curl -p "/api/v1/certificates?name=${cert_name}" | jq -r '.certificates[0].id')"
     credhub curl -p "/api/v1/certificates/${cert_id}/regenerate" -d '{"set_as_transitional": true}' -X POST
done

# BOSH re-deploy, 2 CAs are now trusted

# Step 2: Swap transitional flag
for cert_name in ${cert_names} ; do
     info="$(credhub curl -p "/api/v1/certificates?name=${cert_name}" | jq -c '{ca: .certificates[0].id, version: first(.certificates[0].versions[]|select(.transitional|not)|.id)}')"
     credhub curl -p "/api/v1/certificates/$(echo "${info}" | jq -r .ca)/update_transitional_version" -d '{"version": "'"$(echo "${info}" | jq -r .version)"'"}' -X PUT
     echo "${info}"
done

# Step 3: Re-generate leaf certificates

# neither of these are leaf certificates, so we do not currently re-regenerate as both were re-generated as part of Step 1.

# BOSH re-deploy

# Step 4: Remove transitional flag.
for cert_name in ${cert_names} ; do
     cert_id="$(credhub curl -p "/api/v1/certificates?name=${cert_name}" | jq -r '.certificates[0].id')"
     credhub curl -p "/api/v1/certificates/${cert_id}/update_transitional_version" -d '{"version": null}' -X PUT
done

# BOSH re-deploy
```

After doing the above, both certificates have been re-generated. However they were both re-generated during Step 1, where the new intermediate certificate is still signed using the old root CA.

So while at the end of this all certificates are replaced, when we look at the CA field on the intermediate certificate, we see that it is in fact signed by the *older* version of the root CA (and not validated by clients beyond the expiration date of that old CA).

Reading your comments above reveals the issue:

> That's the reason why it needs to be a CA; but from an outside (i.e. operator) point of view, it's better to think of it as a "normal" certificate.

In this particular context - you are right. However the full rotation scripts we have are looking at attributes returned by `credhub` (rather than specific certificate names) and for step 1 we select using:

```bash
echo "Setting transitional flag for all CA certs..."
for ca in $(credhub curl --path=/api/v1/certificates | jq -r '.certificates[]|select(.versions[]|(.certificate_authority and .generated))|.id' | sort -u) ; do
  credhub curl -p "/api/v1/certificates/${ca}/regenerate" -d '{"set_as_transitional": true}' -X POST >/dev/null
done
```

I think we need to modify this to only select self-signed during step 1, and then during 3 where, we currently select with:

```bash
echo "now re-signing all certs..."
for cert in $(credhub curl --path=/api/v1/certificates | jq -r '.certificates[]|select(.versions[]|((.certificate_authority|not) and .generated))|.id' | sort -u) ; do
  credhub curl -p "/api/v1/certificates/${cert}/regenerate" -d '{}' -X POST >/dev/null
done
```

we should find anything that is generated but not self-signed (instead of excluding certificate authorities).

And that I think that would correctly solve for this particular deployment.

PR #999, opened 2022-10-13
Add Ops-File To Enable CC V2 API Rate Limit

johha:

- Refactor UAA client scopes to use a yaml list instead of single line strings. This will make it easier for ops files to manipulate scopes. This is allowed in uaa-release: https://github.com/cloudfoundry/uaa-release/pull/89
- Add new ops file `enable-cc-v2-rate-limiting.yml` to enable rate limiting for CC V2 API. The ops file is optional and not used per default. Same as for `enable-cc-rate-limiting.yml`

- [ ] YES
- [x] NO

- [x] NO

- Refactor UAA client scopes to use a yaml list instead of single line strings. **This can be a breaking change if cf-d consumers have ops-files which manipulate UAA scopes**
- Add new ops file `enable-cc-v2-rate-limiting.yml` to enable rate limiting for CC V2 API

- [x] YES - new ops file `enable-cc-v2-rate-limiting.yml` to enable rate limiting for CC V2 API
- [ ] NO

No.

UAA should be deployed successfully and login with CF CLI should work

@jochenehret @philippthun @stephanme

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -699,8 +699,8 @@ instance_groups:
             authorities: clients.read,clients.write,clients.admin
             authorized-grant-types: client_credentials
             scopes:
-              - openid
-              - cloud_controller_service_permissions.read
+            - openid
+            - cloud_controller_service_permissions.read
             secret: "((uaa_clients_cc-service-dashboards_secret))"
           cc_service_key_client:
             authorities: credhub.read,credhub.write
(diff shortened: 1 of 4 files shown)


cf-deployment.yml
```diff
@@ -698,7 +698,9 @@ instance_groups:
           cc-service-dashboards:
             authorities: clients.read,clients.write,clients.admin
             authorized-grant-types: client_credentials
-            scope: openid,cloud_controller_service_permissions.read
+            scopes:
+              - openid
```
ctlong, line 702: It looks like a lot of our other yaml lists don't indent the hyphen. It's purely semantic, but let's try and keep it consistent wherever possible :).

```
scopes:
- openid # prefer this
  - openid # over this
```

ctlong (approved):
LGTM

johha:
It looks like `scopes` overrides `scope`. This means if cf-d consumers already have custom ops-files which manipulate `scope` they might run into issues:
```
if !client['scopes'].nil?
  client_data.delete('scopes')
  if client['scopes'].is_a? Array
    client_data['scope'] = client['scopes'].join(',')
  end
```
https://github.com/cloudfoundry/uaa-release/blob/develop/jobs/uaa/templates/config/uaa.yml.erb#L371-L375

From my point of view a yaml list is much better than a single string but it should be mentioned in the release notes that consumers might need to make changes to their ops-files.

PR #1005, opened 2022-11-04
Adapt gorouter ca_cert list to a list

peanball:
## Please take a moment to review the questions before submitting the PR

🚫 We only accept PRs to develop branch. If this is an exception, please specify why 🚫

### WHAT is this change about?

https://github.com/cloudfoundry/gorouter/pull/316 changes the CA list for gorouter (`router.ca_certs`) to an array. The intention is that then the CA list can be augmented via BOSH opsfile, as opsfiles can modify arrays but not concatenate strings.

### What customer problem is being addressed? Use customer persona to define the problem e.g. Alana is unable to...

We are unable to extend the router CA list with some additional custom CAs without repeating the CA list from the CF deployment, as opsfiles cannot concatenate strings.

Each entry in the array may still contain a string that includes multiple contatenated PEM certificates.

### Please provide any contextual information.

https://github.com/cloudfoundry/gorouter/pull/316
https://github.com/cloudfoundry/routing-release/pull/297

### Has a cf-deployment including this change passed [cf-acceptance-tests](https://github.com/cloudfoundry/cf-acceptance-tests)?

- [x] YES
- [ ] NO

### Does this PR introduce a breaking change? Please take a moment to read through the examples before answering the question.

- [x] YES - please choose the category from below. Feel free to provide additional details.
- [ ] NO

> **Types of breaking changes:**

4. modifies the name or deletes a property of a job or instance group in the main manifest

The type of a property is modified. `router.ca_certs` is changed from a `string` to an `array[string]`.

### How should this change be described in cf-deployment release notes?

Change the gorouter CA Certs property to accept an array of PEM encoded certificates, which allow extension via BOSH opsfiles.

### Does this PR introduce a new BOSH release into the base cf-deployment.yml manifest or any ops-files?

- [ ] YES - please specify
- [X] NO

### Does this PR make a change to an experimental or GA'd feature/component?

- [ ] experimental feature/component
- [ ] GA'd feature/component

### Please provide Acceptance Criteria for this change?

When using the routing release that includes https://github.com/cloudfoundry/routing-release/pull/297 and https://github.com/cloudfoundry/gorouter/pull/316, the router.ca_certs property needs to be changed to an array of PEM encoded certs. No change besides the cf-deployment default manifest adapted to the new routing-release is necessary.

### What is the level of urgency for publishing this change?

- [ ] **Urgent** - unblocks current or future work
- [ ] **Slightly Less than Urgent**

PR #1019, opened 2022-11-23
Add a temporary test ops file for routing update

peanball:
In order to update routing release we need a temporary ops file that will change the gorouter ca certs into an array.

Linked PRs:
* https://github.com/cloudfoundry/cf-deployment/pull/1005
* https://github.com/cloudfoundry/routing-release/pull/297
* https://github.com/cloudfoundry/gorouter/pull/316

### DIFF
--- a/ci/input/inputs.yml
+++ b/ci/input/inputs.yml
@@ -29,6 +29,10 @@ baseReleases:
   repository: cloudfoundry-incubator/pxc-release
 - name: routing
   repository: cloudfoundry/routing-release
+  requiredOpsFiles:
+  # this file is only required until routing-release and its breaking config change are consumed
+  # (PR https://github.com/cloudfoundry/cf-deployment/pull/1005). It can be removed afterwards.
+  - operations/test/temp-gorouter-certs-as-list.yml
 - name: silk
   repository: cloudfoundry/silk-release
 - name: statsd-injector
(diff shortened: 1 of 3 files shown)


ci/input/inputs.yml
```diff
@@ -29,6 +29,10 @@ baseReleases:
   repository: cloudfoundry-incubator/pxc-release
 - name: routing
   repository: cloudfoundry/routing-release
+  requiredOpsFiles:
+  # this file is only required until routing-release and its breaking config change are consumed
+  # (PR https://github.com/cloudfoundry/cf-deployment/pull/1005). It can be removed afterwards.
+  - operations/test/temp-gorouter-certs-as-list.yml
```
ctlong, line 35: This will fail `ytt` interpolation. The tool uses `#` to indicate command lines or vars, and comments need to be indicated with `#!`.

ctlong (commented):
I don't think this PR should touch `ci/input/inputs.yml`.

It was a good idea and I appreciate your being fully considerate in your PR :). Unfortunately, the current update-releases template does not look for, or use, the `requiredOpsFiles` field in the `baseReleases` list. That's probably not something we want to change, as base releases are inherently meant to work with vanilla cf-d. That means that once this PR is merged we'd instead hack the generated `ci/pipelines/update-releases.yml` to include this new test ops file in the `update-routing/deploy-cf` task, then manually fly that change. Given that, changing `ci/input/inputs.yml` in this PR doesn't do much, and it's just another change that will have to be undone.

jochenehret:
Ok let's discuss this in tomorrow's WG meeting to find a solution. Indeed, the "requiredOpsFiles" key will not have any effect.

jochenehret:
PR is obsolete, going for approach (3) as discussed above.

PR #574, opened 2018-07-27
Add ops-file for enabling TLS between Gorouter and CC

tcdowney:
### What is this change about?

This PR includes a version of the ops-file that the CAPI team uses to enable and test TLS between the Gorouters and Cloud Controller.

We concatenated the `((application_ca.certificate))` CA certificate
  that the "Routing Integrity" functionality requires with the
  `((service_cf_internal_ca.certificate))` CA certificate that CC's TLS
  certs are signed by to make this ops-file compatible with the
  `enable-routing-integrity.yml` ops-file.

### Please provide contextual information.

CAPI has a separate story for this work because we were hoping to get the ball rolling since we're very interested in getting this functionality into cf-deployment by default:
[CAPI Story #159254762](https://www.pivotaltracker.com/story/show/159254762)

We understand that this will be done in [#158981742](https://www.pivotaltracker.com/story/show/158981742) but hope that this PR will provide additional context and make that one easier.

### Has a cf-deployment including this change passed our [cf-acceptance-tests](https://github.com/cloudfoundry/cf-acceptance-tests)?

- [ ] YES 
- [ ] NO

I have not run CATS, but I have pushed several apps and observed that both API traffic and app traffic is encrypted.

You can test this yourselves by deploying an environment like this:
```
bosh deploy ~/workspace/cf-deployment/cf-deployment.yml \
  -v system_domain=$BOSH_LITE_DOMAIN \
  -o ~/workspace/capi-ci/cf-deployment-operations/skip-cert-verify.yml \
  -o ~/workspace/cf-deployment/operations/bosh-lite.yml \
  -o ~/workspace/cf-deployment/operations/use-compiled-releases.yml \
  -o ~/workspace/capi-ci/cf-deployment-operations/use-latest-capi.yml \
  -o ~/workspace/cf-deployment/operations/experimental/enable-routing-integrity.yml \
  -o ~/workspace/cf-deployment/operations/experimental/enable-router-cc-tls.yml
```

Then do the following:
1. Note the IP addresses of the `router` and `diego-cell` instance groups
2. `bosh ssh`ed on to `api/0` and `router/0` in two separate tabs
3. `sudo su -` in both sessions and run `apt install ngrep -y` to install `ngrep`
4. In the `api/0` session run `ngrep -d any host <router-ip>`
5. In the `router/0` session run `ngrep -d any host <diego-cell-ip>`
6. Push an app `cf push static-app`
[...]

PR #1025, opened 2022-12-03
Gorouter uses TLS for public CAPI blobstore route

Gerg:
- TLS everywhere!
- The blobstore TLS port was already used for internal routes.

Change the `blobstore.((system_domain))` route to use TLS between the Gorouter and the internal CAPI blobstore. This will only apply to environments that use the internal blobstore instead of an external (fog) blobstore.

In addition to the added security of using TLS everywhere, Gorouter will prune non-TLS backends after a 120s TTL. If there is an issue with NATs or something else causes Gorouter to miss the blobstore's route registration heartbeat, then the `blobstore.((system_domain))` will not work, preventing users from downloading packages/droplets/etc.

This is (I believe) the last step in a larger effort to make all of Gorouter's backends TLS and eventually remove the ability to register non-TLS backends.

- [ ] YES
- [x] NO

- [x] NO

Gorouter uses TLS for public CAPI blobstore route.

- [ ] YES - please specify
- [x] NO

- [x] GA'd feature/component

1. Deploy using the default, internal capi blobstore (`singleton-blobstore`)
2. Push an app
3. Download its droplet: 
```
❯ cf download-droplet <app>
Downloading current droplet for app <app> in org <org> / space <space> as admin...
Droplet downloaded successfully at /path/to/droplet_blah.tgz
OK
```

@cloudfoundry/wg-app-runtime-interfaces-capi-approvers 
@ameowlia 
@tcdowney who did [the same thing for CC](https://github.com/cloudfoundry/cf-deployment/pull/574)

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -870,8 +870,10 @@ instance_groups:
         - name: blobstore
           port: 8080
           registration_interval: 20s
+          server_cert_domain_san: "blobstore.((system_domain))"
           tags:
             component: blobstore
+          tls_port: 4443
           uris:
           - blobstore.((system_domain))
 - name: api
@@ -2102,6 +2104,7 @@ variables:
     ca: service_cf_internal_ca
     common_name: blobstore.service.cf.internal
     alternative_names:
+    - "blobstore.((system_domain))"
     - blobstore.service.cf.internal
 - name: diego_auctioneer_client
   type: certificate


cf-deployment.yml
```diff
@@ -870,8 +870,10 @@ instance_groups:
         - name: blobstore
           port: 8080
```
ameowlia, line 871: You can remove this now ✨

cf-deployment.yml
```diff
@@ -870,8 +870,10 @@ instance_groups:
...
           registration_interval: 20s
+          server_cert_domain_san: "blobstore.((system_domain))"
           tags:
             component: blobstore
+          tls_port: 4443
```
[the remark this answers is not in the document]
Gerg: It might not be 😓.

I'll double check that we aren't making any authz decisions based on which port is used.

cf-deployment.yml
Gerg: During a past iteration of this effort, there [was concern](https://www.pivotaltracker.com/story/show/164937832/comments/202023498) about removing the non-TLS ports, since they could be cached in some ancient LRPs. 

We can re-evaluate the risk here, since time has continued to progress. cc @tcdowney

ameowlia: I think gorouter will just disregard the `port` param if you provide `tls_port`.
* gorouter makes the endpoint [here](https://github.com/cloudfoundry/gorouter/blob/43d4f29110f9f007b4ce59939ea9add3b537df92/mbus/subscriber.go#L47).
* and it calculated port [here](https://github.com/cloudfoundry/gorouter/blob/43d4f29110f9f007b4ce59939ea9add3b537df92/mbus/subscriber.go#L85-L90) and only returns one or the other.


If you want both endpoints to still exist, you will need to add it as a separate route in the array here.

Gerg: Oh, I see. I thought you were talking about removing the actual port, not just the register configuration.

cf-deployment.yml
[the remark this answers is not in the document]
Gerg: I'll draft this PR while I investigate.

Gerg:
Implemented by https://github.com/cloudfoundry/cf-deployment/commit/f9fbf523fe8b32cd5bd4e1f1302085e9bf49d16d

(With help from https://github.com/cloudfoundry/capi-release/commit/385119fab0eb8a38c5bbbeb05ded70326b198507)

PR #1030, opened 2022-12-21
Check bbr-github-release every 5m

jochenehret:
* default is 1m and we sometimes run into the github unauthenticated rate limit


ctlong (approved):
LGTM! An alternative is to authenticate the request, but for this resource I don't think it matters much.

PR #1032, opened 2023-01-02
2 attempts for "bosh-deploy-cf-develop" tasks

jochenehret:
* if this task fails in "upgrade-deploy" or "experimental-deploy", we have to start again from the beginning which takes very long (because first "bosh-deploy-cf-latest-release" is executed)

### DIFF
--- a/ci/pipelines/cf-deployment.yml
+++ b/ci/pipelines/cf-deployment.yml
@@ -836,6 +836,7 @@ jobs:
         environments/test/trelawney/add-external-ips-for-cloud-sql-db.yml
         environments/test/trelawney/scale-log-api-to-4.yml
   - task: bosh-deploy-cf-develop
+    attempts: 2
     file: cf-deployment-concourse-tasks/bosh-deploy/task.yml
     input_mapping:
       bbl-state: relint-envs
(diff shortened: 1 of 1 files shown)


ctlong (commented):
From a quick scan of the relevant pipeline jobs, it seems to me that the flakey failures for the `upgrade-deploy` and `experimental-deploy` jobs often occur during the `bosh-deploy-cf-latest-release` tasks rather than the `bosh-deploy-cf-develop` tasks.

❓ Shouldn't the number of attempts be increased for the flakey `bosh-deploy-cf-latest-release` tasks instead / as well?

Otherwise I'm fine with this workaround for now. At some point we should probably try and figure out where the flakiness comes from and resolve the root problem, if possible.

jochenehret:
Added 2 attempts for "bosh-deploy-cf-latest-release" tasks. If they fail it's usually a problem with the IaaS provider:
```
Error: Timed out pinging VM 'i-038dbac0fb6ff0de2' with agent 'f447db17-7167-44c3-8a90-c5a9b0f6e39e' after 600 seconds
```

ctlong (approved):

PR #1068, opened 2023-04-07
Feat: Send logs and metrics to Log Cache using mTLS

Benjamintf1:

Aggregate drains now uses mututal tls for log-cache connections

It makes log-cache more secure

- [x] YES
- [ ] NO

- [x] NO

- [ ] YES - please specify
- [x] NO

- [x] GA'd feature/component

Log cache continues to work

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -86,7 +86,6 @@ addons:
         cert: "((syslog_agent_metrics_tls.certificate))"
         key: "((syslog_agent_metrics_tls.private_key))"
         server_name: syslog_agent_metrics
-      drain_ca_cert: "((log_cache_syslog_tls.ca))"
 
 - name: prom_scraper
   include:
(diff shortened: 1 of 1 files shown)


Benjamintf1:
https://github.com/cloudfoundry/loggregator-agent-release/releases/tag/v7.2.0 Ready to merge as soon as this release is in

Benjamintf1:
(Actually might want to check this doesn't cause downtime)

Benjamintf1:
If log-cache is last, then it's no downtime, if it isn't we might want to wait to merge this.

ctlong:
@Benjamintf1 maybe we should stage this change in two PRs: 1) add the certificates to the syslog binding cache, and 2) remove the `drain_ca_cert` property from the syslog agents.

That would eliminate any potential downtime, right?

Benjamintf1:
That's not the problem. The problem is, if we roll log-cache before say, diego-cells/etc, then old syslog-agents(that can't get the *client* information) would start failing. 

The ca cert thing should be fine I believe. At least, mostly. New agents will get the new ca through the new endpoint alongside the client creds. It's if log-cache starts enforcing mtls before the agents have mtls that's the problem.

ctlong:
The deployment order appears to be:

```
$ cat cf-deployment.yml | yq '.instance_groups.[] | .name'
smoke-tests
nats
database
diego-api
uaa
singleton-blobstore
api
cc-worker
scheduler
router
tcp-router
log-cache
doppler
diego-cell
log-api
credhub
rotate-cc-database-key
```

That means that the Syslog Binding Cache (on `scheduler`) should be updated to start serving a Log Cache address with mTLS prior to the Log Cache Syslog Server (on `log-cache`) requiring mTLS. However, that's only a small time gap and I think most Syslog Agents (added on to every instance) only poll the Syslog Binding Cache every 30 seconds, so this could result in some amount of log downtime in Log Cache. It's only on the one upgrade for a small amount of time though, and Log Cache is meant to be transient, so I think this is fine.

ctlong (approved):

Issue #1048, opened 2023-02-10
cflinuxfs4 Rollout - Step 2: Remove deprecated cflinuxfs3 stack


jochenehret:
### What is this issue about?

Details out https://github.com/cloudfoundry/cf-deployment/issues/989.

In step 2 of the rollout, we will remove the deprecated/unmaintained cflinuxfs3 stack and the cflinuxfs3 buildpacks. This should happen latest end of April or beginning of May 2023 when the support for Ubuntu Bionic ends: https://wiki.ubuntu.com/Releases

As this is a breaking change, we would ship it as a major release.

Technically we would remove all cflinuxfs3 components from [cf-deployment.yml](https://github.com/cloudfoundry/cf-deployment/blob/main/cf-deployment.yml). We would provide an ops file to add the cflinuxfs3 components again.

jochenehret:
PR: https://github.com/cloudfoundry/cf-deployment/pull/1078

PR #1078, opened 2023-04-27
Remove CFLINUXFS3 from default cf-deployment.yml

FloThinksPi:

This contribution removes cflinuxfs3 from the default `cf-deployments.yml`. However ops files have been added to revert the new default behaviour.

The oprations file `use-cflinuxfs3.yml` will introduce cflinuxfs3 again. Additionally there are the operation files `use-cflinuxfs3-with-compiled-releases.yml` as well as `use-cflinuxfs3-with-trusted-ca-cert-for-apps.yml`.

Due to a limitation in bosh a replace statement cannot be skipped if e.g. cflinuxfs3 fields are not present. The optional bosh parameter (?) would create them again. Without the parameter the templating fails. Thus the only option seems to
move parts of other ops files that interfere with cflinuxfs3 provided field to own operation files.

This now means when using `use-trusted-ca-cert-for-apps.yml` in combination with `use-cflinuxfs3.yml` one also must use `use-cflinuxfs3-with-trusted-ca-cert-for-apps.yml` to bring back the old behaviour of both ops files. When using just cflinuxfs4 using `use-trusted-ca-cert-for-apps.yml` is sufficient on its own as it also has no reference to cflinuxfs3 anymore. Same applies to the behaviour of the operations file `use-compiled-releases.yml`.

Additionally this contibution adopts tests and pipelines responsible for bumping the cflinuxfs3 stack. The version bump shall only be applied to the operation file `use-cflinuxfs3.yml` as it now brings in the cflinuxfs3 release which is not part of `cf-deployment.yml` anymore by default.

The cflinuxfs3 stack is out of maintenance soon and should not be active anymore by default.

https://github.com/cloudfoundry/cf-deployment/issues/1048

- [ ] YES
- [x] NO

- [X] YES
- [ ] NO

1. **causes app or operator downtime**  -> Possible if apps rely on old and remove cflinuxfs3 stack and operators take no action to bring it back via provided operations file
3. modifies, deletes or moves the name of a job or instance group in the main manifest -> removes the cflinuxfs3 release
4. modifies the name or deletes a property of a job or instance group in the main manifest -> removes the cflinuxfs3 release
6. requires out-of-band manual intervention on the part of the operator -> Apps running on the cflinuxfs3 stack willl be stuck in a crashed state once the stack is removed. An operator needs to make sure all apps on the foundation are migrated to the cflinuxfs4 stack first.
7. modifies the ops-file path, changes the type, changes the values or removes ops-files from the following folders
   - operations

> Removed cflinuxfs3 stack from default cf-deployment.yml, added ops files to bring the (unsupported)stack back.

- [ ] YES - please specify
- [X] NO

- [X] GA'd feature/component

@stephanme , @jochenehret , @johha, @philippthun


jochenehret (changes requested):
The [inputs.yml](https://github.com/cloudfoundry/cf-deployment/blob/main/ci/input/inputs.yml) file must be updated. You must move the "cflinuxfs3" release from "baseReleases" to "opsReleases" and regenerate the "update-releases.yml" pipeline definition with the [configure script](https://github.com/cloudfoundry/cf-deployment/blob/main/ci/configure).

jochenehret:
The ops files for the "deploy-cf" job in the "bionic-stemcell-validation" pipeline must be updated:
https://github.com/cloudfoundry/cf-deployment/blob/77e177eebc0ca6649e5a0ab2c68373f3991d502f/ci/pipelines/bionic-stemcell.yml#L85
The pipeline can be re-uploaded when the changes are on the "release-candidate" branch.

jochenehret (dismissed):

PR #1079, opened 2023-04-27
Add JWT certificate and signing alg

strehle:

New UAA key parameters for JWT:
signingAlg  -> https://tools.ietf.org/html/rfc7518#section-3.1 
signingCert -> https://tools.ietf.org/html/rfc7517#section-4.7

Acceptance UAA should have set this parameters to activate new behaviour.

no problem, but new feature

JWT token creation , see https://tools.ietf.org/html/rfc7518#section-3.1  and https://tools.ietf.org/html/rfc7517#section-4.7

- [X] YES
- [ ] NO

- [X] NO

>    - `./addons`
>

- [ ] YES - please specify
- [X] NO

- [X] GA'd feature/component

New UAA in /token_keys show xt5 claim with the certificate as trust anchor , e.g. https://login.uaa-acceptance.cf-app.com/token_keys

> _Few examples_

UAA -> https://github.com/cloudfoundry/uaa


jochenehret:
Is this really no breaking change? This PR basically changes the type of a BOSH variable (from "rsa" to "certificate"). What does this mean for existing deployments? Can BOSH update the credential? Do we need any intermediate steps to update the UAA instances?

strehle:
> Is this really no breaking change? This PR basically changes the type of a BOSH variable (from "rsa" to "certificate"). What does this mean for existing deployments? Can BOSH update the credential? Do we need any intermediate steps to update the UAA instances?

Hi , the uaa_jwt_signing_key should be a certificate yes similar to SAML config. The private key is the only required config; the cert is optional for UAA Startup therefore I see this change as non breaking.

jochenehret:
Ok so from UAA perspective this would not be a breaking change. But what about BOSH and CredHub? If someone has already a "rsa" credential in CredHub and now we change it to "certificate", will BOSH automatically change the type?

strehle:
> Ok so from UAA perspective this would not be a breaking change. But what about BOSH and CredHub? If someone has already a "rsa" credential in CredHub and now we change it to "certificate", will BOSH automatically change the type?

@bruce-ricard @peterhaochen47 can you help about credhub behavior? I found https://github.com/cloudfoundry/credhub/blob/main/docs/credential-types.md , so from my understanding change from rsa to certificate is no breaking change, because field private_key is in both types available and this is what UAA really needs . The existence of certificate is optional.

ae-govau:
Hmm, this changes something that was previously a key to something that is now a certificate. Amongst other things, certificates have expiration dates...

What does BOSH do when it sees a variable declared that exists but is the wrong type? I bet it's not documented. Is it different if credhub is used? Is that documented?

I worry that at best this change is brittle and it's not clear what benefit it brings? Section 4.7 referenced above is an optional field - it's not required to be returned.

davewalter:
Would it make sense to simply change the name of the variable as part of this PR? This should cause BOSH to generate a new certificate.

strehle:
> Reference

please see https://github.com/cloudfoundry/cf-deployment/pull/1079/commits/a0f6421577fa854d7c44f7aeedef5609013ed13d so with that we have 2 keys in OIDC but key-1 is not touched at all. Key2 is created with the X509 enhancements

x5t in the JSON Web Keys is something which is used by Microsoft but its usage is in standard defined. We need it in UAA to perform OIDC calls towards Azure AD without secrets.
Otherwise UAA would have to store a secret from Azure AD in UAA DB. The client authentication mechanism https://oauth.net/private-key-jwt/ allows to omit secrets. The x5t from UAA needs to be stored in Azure AD, then Azure trusts a UAA. If UAA has the certificate we add the x5t, see others approach https://github.com/OpenIDC/mod_auth_openidc/discussions/762

finally, MS documentation https://learn.microsoft.com/en-us/azure/active-directory/develop/certificate-credentials .

strehle:
> I worry that at best this change is brittle and it's not clear what benefit it brings? Section 4.7 referenced above is an optional field - it's not required to be returned.

You can say that the big companies are brittle, yes, but is part of the standard. If you want omit secrets and integrate Azure AD then you have to do this.
I want to show others with that what UAA is able to do. If you customize your bosh deployment you can do this already on your own. The OIDC integration has the advantage, that you can replace SAML and LDAP because OIDC allows the browser and password (API) flows and you can link your UAA to your corporate Azure.

jochenehret:
Using a new credential should cause less problems for the transition. We'll discuss this PR in tomorrow's ARD WG meeting.

jochenehret (approved):

PR #1098, opened 2023-06-14
Fix cflinuxfs4-compat-release for fresh-deploy

jochenehret:

Fix "fresh-deploy" validation setup. The test ops file `add-persistent-isolation-segment-diego-cell.yml` uses the "cflinuxfs4" release, so we must overwrite that with another ops file.

https://github.com/cloudfoundry/cflinuxfs4-compat-release/issues/1

https://github.com/cloudfoundry/cflinuxfs4-compat-release/issues/1

- [ ] YES
- [x] NO

- [x] NO

Not relevant for release notes.

- [ ] YES - please specify
- [x] NO

- [ ] GA'd feature/component

CI job https://concourse.wg-ard.ci.cloudfoundry.org/teams/main/pipelines/cf-deployment/jobs/fresh-deploy must pass.

### DIFF
--- a/ci/pipelines/cf-deployment.yml
+++ b/ci/pipelines/cf-deployment.yml
@@ -528,6 +528,7 @@ jobs:
         operations/rename-network-and-deployment.yml
         operations/set-bbs-active-key.yml
         operations/test/add-persistent-isolation-segment-diego-cell.yml
+        operations/test/use-cflinuxfs4-compat-isolation-segment-diego-cell.yml
         operations/test/add-persistent-isolation-segment-router.yml
         operations/rename-isolation-segment-network.yml
         operations/addons/enable-component-syslog.yml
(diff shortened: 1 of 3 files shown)


ctlong:
@jochenehret seems like the isolated-diego-cell needs to be told to use cflinuxfs4-compat as well. See this CI run: https://concourse.wg-ard.ci.cloudfoundry.org/teams/main/pipelines/cf-deployment/jobs/fresh-deploy/builds/343.

jochenehret:
@ctlong : I think it's better to replace the cflinuxfs4 release. A BOSH deployment should not contain releases that are not really in use. Technically it's possible to keep both, that is no problem from a stack perspective. We (SAP) and possibly others generate a bill of material that lists all consumed artifacts. Here it would also be preferable to only list what is really used.

We'll check if we can simplify this PR somehow.

PR #1114, opened 2023-08-22
ci(pull-requests): Don't require reviews to fail PRs made against main

ctlong:

We currently only use these PR resources to error if the main branch is the base branch of the PR. Since presence of a PR to main is all that we care about, and the contents of the PR are never examined, let alone run, we should be able to disregard the requirements for reviews in this case.

The two CI jobs using these resources exist for the sole purpose of alerting PR creators and approvers that a PR has been incorrectly made against main instead of develop. Currently, a PR approval is required to trigger the two jobs, which means that it's initially on approvers to catch the incorrect base branch. I've missed that mistake two or three times in the past two weeks, and it's been causing me frustration because, after approving a PR, I then have to go back and explain that the base branch is wrong, which usually requires a rebase to fix, which requires a new PR approval...

Ever since we changed the default branch of CATs and CF-D from develop to main (for branch protection reasons IIRC), probably 50% of new PRs have been accidentally made against main. That's despite clear messages in the PR templates to only create PRs against develop.

- [ ] YES
- [x] NO - just a CI change

- [x] NO

- [ ] YES - please specify
- [x] NO

- [ ] GA'd feature/component

When a PR is made against CF-D or CATs' `main` branch, the fail-prs-to-cats-main / fail-prs-to-cf-deployment-main job will immediately run and mark the PR as failing this check.

None

### DIFF
--- a/ci/pipelines/pull-requests.yml
+++ b/ci/pipelines/pull-requests.yml
@@ -37,7 +37,6 @@ resources:
       access_token: ((ard_wg_gitbot_token))
       repository: cloudfoundry/cf-deployment
       base_branch: main
-      required_review_approvals: 1
   - name: cf-deployment-rc
     type: pull-request
     source:
@@ -65,7 +64,6 @@ resources:
       access_token: ((ard_wg_gitbot_token))
       repository: cloudfoundry/cf-acceptance-tests
       base_branch: main
-      required_review_approvals: 1
   - name: cats-develop
     type: pull-request
     source:


davewalter:
Is the approval requirement currently causing problems? I am concerned about the possibility of a PR accidentally being merged if it is removed.

chombium:
@ctlong, I agree with @davewalter. Tbh, I don't know what's the process for merging develop in main. If there is some kind of automation, some kind of ci/cd pipeline that checks everything is all right and can be merged and the only way of merging things is through this pipeline, than we might do that, but I still think that it doesn't hurt if someone takes a look at the changes.

ctlong:
@davewalter @chombium to be clear this does not change the majority of CF-D and CATs PR pipelines, and in fact only impacts the following two jobs:
* [fail-prs-to-cf-deployment-main](https://concourse.wg-ard.ci.cloudfoundry.org/teams/main/pipelines/pull-requests/jobs/fail-prs-to-cf-deployment-main/builds/2)
* [fail-prs-to-cats-main](https://concourse.wg-ard.ci.cloudfoundry.org/teams/main/pipelines/pull-requests/jobs/fail-prs-to-cats-main/builds/20)

Ever since we changed the default branch of CATs and CF-D from `develop` to `main` (for branch protection reasons IIRC), probably 50% of new PRs have been accidentally made against `main`. That's despite clear messages in the PR templates to only create PRs against `develop`.

The two jobs that I mentioned above exist for the sole purpose of alerting PR creators and approvers that a PR has been incorrectly made against `main` instead of `develop`. Currently, a PR approval is required to trigger the two jobs, which means that it's initially on approvers to catch the incorrect base branch. I've missed that mistake two or three times in the past two weeks, and it's been causing me frustration because, after approving a PR, I then have to go back and explain that the base branch is wrong, which usually requires a rebase to fix, which requires a new PR approval...

That's not to mention the risk that an approver merges a PR before the jobs above mark it as incorrect, not realizing it's made against `main` until it's too late. This is a more annoying problem to fix than if unit tests aren't passing. Also, the security risks that caused us to initially require approvals to run PRs through CI shouldn't apply here, because these two jobs don't actually run, or interact with, the PR code.

chombium:
@ctlong thanks for the detailed clarification. I really like this idea and we should do it to the other repos as well. 

I've always hated it that when I open a PR in other CF repos, I have to wait for an approver to let the PR validation run. Sometimes I've waited at least a day because of shifted time zones and working hours, and sometimes even more. I really like the idea when someone opens a PR to be immediately informed about the state of the PR.

PR #1144, opened 2023-12-11
Create smoke-test-timeout-scale.yml

ramonskie:
if ephemeral ip not enabled within gcp a timeout will cause the smoke tests to fail see


ctlong (changes requested):
Can you please update the relevant ops file README?

Also, in order for this PR to pass our CI, the [relevant unit tests](https://github.com/cloudfoundry/cf-deployment/blob/main/units/tests/test_test/operations.yml) are going to need to be updated.

ctlong (approved):

PR #1145, opened 2023-12-12
Use otel-collector-release for otel-collector

mkocher:
Also changing sample vars file path to be compatible with both windows and linux jobs

Moving otel-collector out to its own release.

Other deployments may wish to use otel-collector without using loggregator-agent.

- [ ] YES
- [X] NO

- [X] NO

otel-collector now comes from the otel-collector-release

- [X] YES - otel-collector-release
- [ ] NO

- [ ] GA'd feature/component

`bosh releases` should show otel-collector-release is deployed

@ctlong

### DIFF
--- a/operations/experimental/example-vars-files/vars-override-otel-collector-exporters.yml
+++ b/operations/experimental/example-vars-files/vars-override-otel-collector-exporters.yml
@@ -1,6 +1,6 @@
 ---
 otel_collector_metric_exporters:
   file/test:
-    path: /var/vcap/sys/log/otel-collector/file.log
+    path: /tmp/otel-collector-file.log
   otlp/test:
     endpoint: otelcol:4317
--- a/ci/input/inputs.yml
+++ b/ci/input/inputs.yml
@@ -25,6 +25,8 @@ baseReleases:
   repository: cloudfoundry/loggregator-agent-release
 - name: nats
   repository: cloudfoundry/nats-release
+- name: otel-collector
+  repository: cloudfoundry/otel-collector-release
 - name: pxc
   repository: cloudfoundry/pxc-release
 - name: routing
--- a/ci/pipelines/update-releases.yml
+++ b/ci/pipelines/update-releases.yml
@@ -36,6 +36,8 @@ groups:
   - update-loggregator-agent
   - acquire-nats-pre-dev-lock
   - update-nats
+  - acquire-otel-collector-pre-dev-lock
+  - update-otel-collector
   - acquire-pxc-pre-dev-lock
   - update-pxc
   - acquire-routing-pre-dev-lock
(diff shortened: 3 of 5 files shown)


jochenehret:
This PR introduces a new BOSH release: https://bosh.io/releases/github.com/cloudfoundry/otel-collector-release?all=1
Can you please update https://github.com/cloudfoundry/cf-deployment/blob/main/ci/input/inputs.yml with the new release? Then we can re-generate the [update-releases pipeline](https://concourse.wg-ard.ci.cloudfoundry.org/teams/main/pipelines/update-releases). This pipeline automatically checks for new BOSH release versions from bosh.io and updates the ops file.

mkocher:
@jochenehret thanks, we updated inputs.yml and cleared up our CLA signing

operations/experimental/example-vars-files/vars-override-otel-collector-exporters.yml
```diff
@@ -1,6 +1,6 @@
 ---
 otel_collector_metric_exporters:
   file/test:
-    path: /var/vcap/sys/log/otel-collector/file.log
+    path: /tmp/otel-collector-file.log
```
chombium, line 4: I know that this is an experimental release and test config, so maybe therefore you have changed the path to the log file.
I don't think that this is good because of the following:
- All BOSH deployments write their logs in `/var/vcap/log/<job-name>`. If the path is changed it will be confusing for the users
- If the logs is written in `/tmp` the logs will not be persisted in-between restarts of the VM as the `/tmp`. This would make troubleshooting harder
- The filename should also follow the BOSH conventions `<jobname>.stdout.log`, `<jobname>.stderr.log` and it would be good if the custom logs have meaningful names as well

chombium (changes requested):
It looks good, but I think we shouldn't adjust the path to the log file.

operations/experimental/example-vars-files/vars-override-otel-collector-exporters.yml
mkocher: We made this change so that in CI the log files for both otel-collector and otel-collector-windows can go to the same place, so we can deploy them with the same variable. On windows the /var/vcap/sys/log/otel-collector/ path doesn't exist. In general we don't expect users to use the file exporter except for testing.

Some Options:
1. We could use a different vars file in CI, however that has the downside of leaving this vars file untested in the logging and metrics pipelines
1. We could use two variables, but this has the downside of not sharing the exporter config across all the collectors
1. Leave it as is, and expect users to not copy this vars file verbatim

What do you think?

chombium: Thank you very much for the detailed  explanation @mkocher. I guess it's ok if we move the log file to `/tmp`. The component is in early stage and still experimental. We would have to think about how to configure it properly before going GA.

PR #1155, opened 2024-01-18
Local blobstore now supports TLS

sethboyles:

This limits the singleton-blobstore to only use TLS for all endpoints.

Better security on singleton-blobstore droplet & package upload/download endpoints.

This PR requires that https://github.com/cloudfoundry/capi-release/pull/377, which is in capi-release 1.171.0. This is already merged into cf-d!

- [x] YES
- [ ] NO

- [X] NO

I'm pretty sure it doesn't qualify for 7 as a breaking change? It removes an additional variable from the manifest, but I think that ops file is only meant to work with the same version of cf-deployment.

>    - `./addons`

> Singleton Blobstore now only uses TLS

- [ ] YES - please specify
- [X] NO

- [X] GA'd feature/component

There should be no noticeable behavioral change.  This port will now be used for package/droplet uploads and downloads and will be encrypted on the blobstore vm.

@dalvarado 

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -816,7 +816,8 @@ instance_groups:
       route_registrar:
         routes:
         - name: blobstore
-          port: 8080
+          tls_port: 8081
+          server_cert_domain_san: "blobstore.((system_domain))"
           registration_interval: 20s
           tags:
             component: blobstore


johha (approved):
LGTM

PR #1175, opened 2024-05-07
Remove unneeded pipelines

jochenehret:

Remove unneeded pipelines. The "Jammy" stemcell validation was integrated into the "cf-deployment" pipeline quite some time ago. The "acceptance.yml" pipeline is not used at all.

Alana does not want to have outdated pipeline definitions in the repository.


dimivel (approved):
LGTM

ctlong (approved):

PR #1191, opened 2024-07-17
Update tcp-router to run serially

geofffranks:

Updates tcp-router VMs to deploy serially, immediately after gorouters, so that they can finish being updated before the parallel block containing diego-cells run.

This allows us to ensure that all tcp-router related code + features are deployed and configured prior to updating diego-cells with code requesting those features via route-emitter.

We want tcp-router to deploy serially immediately after gorouter. This is a required pre-requisite before we can have users do a one stage deploy with no downtime to turn on a future TCP-router feature.

Currently during a deploy the tcp-router is updated in parallel to diego-cells. We think it deploys in this order because TCP-router is often an after thought and it was just shoved in at the end. 

- [x] YES
- [ ] NO

- [x] NO

Not sure if this is needed? But "tcp-router now updates serially after gorouter, instead of in parallel to diego-cells"

- [ ] YES - please specify
- [x] NO

- [x] GA'd feature/component

- Fresh deploys of cf-deployment succeed (we tested this)
- Upgrades to cf-deployment succeed with no change to uptimer results (we tested this)
- Upgrades to cf-deployment when using ops files for single AZ + fast deploys with downtime + danger succeed (we tested this)

@cloudfoundry/wg-app-runtime-platform-networking-approvers

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -1444,6 +1444,8 @@ instance_groups:
   instances: 2
   vm_type: minimal
   stemcell: default
+  update:
+    serial: true
   vm_extensions:
   - cf-tcp-router-network-properties
   networks:


geofffranks:
We have validated that deploys from scratch work without issue, and that deploys that recreate all vms result in no new downtime to tcp routes. When testing against our branch that adds graceful LB drain/recovery support to tcp router, we see no downtime to tcp routes.

dimivel (approved):

jochenehret:
Shouldn't we remove `update: serial` from the `router` group? Then `router` and `tcp-router` could be updated in parallel.

jochenehret (approved):

PR #1200, opened 2024-09-04
Add consumes/provides for mapfs

winkingturtle-vmw:

Adding a consumes/provides will make it possible to consume this release from another job. We need to add the same job in an isolation-segment which requires us to have this change.

Alana is unable to use the same in an isolation-segment

- [x] YES
- [ ] NO

- [x] NO

>    - `./addons`
>

- [ ] YES - please specify
- [x] NO

- [x] GA'd feature/component

> _Few examples_

@ameowlia

### DIFF
--- a/operations/enable-nfs-volume-service.yml
+++ b/operations/enable-nfs-volume-service.yml
@@ -85,11 +85,15 @@
           server_cert: ((nfsv3driver_cert.certificate))
           server_key: ((nfsv3driver_cert.private_key))
     release: nfs-volume
+    consumes:
+      mapfs: {from: mapfs-diego-cell}
 - type: replace
   path: /instance_groups/name=diego-cell/jobs/name=mapfs?
   value:
     name: mapfs
     release: mapfs
+    provides:
+      mapfs: {as: mapfs-diego-cell}
 - type: replace
   path: /variables/-
   value:


winkingturtle-vmw:
Requesting review from @cloudfoundry/wg-app-runtime-deployments-cf-deployment-approvers

davewalter:
@winkingturtle-vmw Could you please provide more detail about the upcoming change to isolation segments that requires this change? Would it be possible for this change to be included in the ops-file that adds the isolation segment?

ameowlia:
👋 Hi @davewalter, 

We have been gifted responsibility for the volume service releases. In the nfs-volume-service-release pipeline we need to run tests against two different nfs configurations (ldap and non-ldap). In order to only use one environment, we deploy the non-ldap configuration on the normal CF diego cells, and we deploy the ldap config on an iso-seg. We are using the isolation segment configs from cf-d, so the isolated cells are in the same bosh deployment as the normal cells. 

In this deployment setup there is a mapfs job deloyed on two different instance groups. Bosh links doesn't like this, so we need to define explicit provides/consumes links.

We could maintain our own opsfile that does this, but we would like to minimize the amount of things we own, so we hoped that we could make the change here. This would not change the behavior for anyone currently using this opsfile. This change would allow users like us to deploy volume services on isolation-segments as well.

davewalter:
Hi @ameowlia, 👋 

Thanks for the context. I can definitely appreciate your desire to not have to maintain a custom ops-file to apply this change, but my concern is that it is only required for testing and I am worried that it would change how users would consume the mapfs link in the (admittedly unlikely) event that they have a release that also makes use of it.

ameowlia:
👍 fair, fair. I'll close this and we will maintain our own opsfile.

PR #1201, opened 2024-09-06
Add missing ldap_ca_cert

winkingturtle-vmw:

Fix nfs ldap opsfile to include ldap_ca_cert

Current opsfile is broken

- [ ] YES 
- [x] NO (There is not cats test that covers this opsfile)

- [x] NO

As an operator I can deploy nfs with ldap and I can then cf push my app

- [ ] YES - please specify
- [x] NO

- [x] GA'd feature/component

Using this opsfile will work.

### DIFF
--- a/operations/enable-nfs-ldap.yml
+++ b/operations/enable-nfs-ldap.yml
@@ -21,5 +21,8 @@
   path: /instance_groups/name=diego-cell/jobs/name=nfsv3driver/properties/nfsv3driver/allowed-in-source?
   value: ""
 - type: replace
+  path: /instance_groups/name=diego-cell/jobs/name=nfsv3driver/properties/nfsv3driver/ldap_ca_cert?
+  value: ((ldap_test_server_ssl.ca))
+- type: replace
   path: /instance_groups/name=nfs-broker-push/jobs/name=nfsbrokerpush/properties/nfsbrokerpush/ldap_enabled?
   value: true


operations/enable-nfs-ldap.yml
```diff
@@ -20,6 +20,9 @@
 - type: replace
   path: /instance_groups/name=diego-cell/jobs/name=nfsv3driver/properties/nfsv3driver/allowed-in-source?
   value: ""
+- type: replace
+  path: /instance_groups/name=diego-cell/jobs/name=nfsv3driver/properties/nfsv3driver/ldap_ca_cert?
+  value: ((ldap_test_server_ssl.ca))
```
jochenehret, line 25: The certificate `ldap_test_server_ssl` is defined in [enable-nfs-test-ldapserver.yml](./operations/test/enable-nfs-test-ldapserver.yml), but that ops file is only for test purposes. Your intention is to make the ldap's SSL CA certificate configurable, right? Should we use another name for the CA (without "test")? You should also enhance the [vars-enable-nfs-ldap.yml](./operations/example-vars-files/vars-enable-nfs-ldap.yml) file to include the CA. See e.g. [vars-use-trusted-ca-cert-for-apps.yml](./operations/example-vars-files/vars-use-trusted-ca-cert-for-apps.yml).

jochenehret (changes requested):

operations/enable-nfs-ldap.yml
winkingturtle-vmw: @jochenehret I have added an example value for the CA and updated the PR.

I don't have a strong opinions about the naming convention, since you can't use `enable-nfs-ldap.yml` without `enabled-nfs-test-ldapserver.yml`. Even if your ldap server is external to the foundation, you'd have to have it deployed with a cert.

jochenehret: You should be able to use `enable-nfs-ldap.yml` without `enable-nfs-test-ldapserver.yml`. The latter ops file deploys a test ldap server and also generates a test SSL certificate. For a productive setup, you just need `enable-nfs-ldap.yml` and provide a SSL certificate. That certificate would have to be generated externally (not BOSH managed).

operations/test/enable-nfs-test-ldapserver.yml
jochenehret, line 1: There is no need to change anything in this file. If you apply this ops file after `enable-nfs-ldap.yml`, you get the desired result:
```
  - name: nfsv3driver
    properties:
      nfsv3driver:
        allowed-in-source: ""
        ldap_ca_cert: ((ldap_test_server_ssl.ca))
```
Renaming `ldap_test_server_ssl` to `ldap_ca_cert` would be highly confusing because `ldap_test_server_ssl` is not a certificate authority (but `ldap_test_server_ssl.ca` refers to its CA).

jochenehret (changes requested):

operations/test/enable-nfs-test-ldapserver.yml
winkingturtle-vmw: I am not sure I follow what you are saying. We wouldn't want to apply yet another opsfile in order to get the basic happy path working. The reason for this PR is to make sure [SMB and NFS instruction](https://github.com/cloudfoundry/nfs-volume-release/blob/develop/docs/02-deploying-nfs-volume-service.md) are the same and a person can deploy both in the same way to have a fully working test-server and binding. 

If your suggestion is to define another opsfile somewhere else outside of Cf-deployment, we can leave this functionality as is and close this PR instead.

winkingturtle-vmw:
```bash
bosh interpolate \
 -o ./operations/enable-nfs-volume-service.yml \
 -o ./operations/enable-nfs-ldap.yml \
 -o ./operations/test/enable-nfs-test-server.yml \
 -o ./operations/test/enable-nfs-test-ldapserver.yml \
 cf-deployment.yml
 ```

This is the option that I want to work out of the box. Enable ldap for diego-cell and ensure that there is a test-server and test-ldap-server to talk to. 

> bosh interpolate \
 -o ./operations/enable-nfs-volume-service.yml \
 -o ./operations/test/enable-nfs-test-server.yml \
 -o ./operations/test/enable-nfs-test-ldapserver.yml \
 cf-deployment.yml
Note that you don't need the enable-nfs-ldap.yml ops file as no external hosts or ports need to be specified.

This is not true, you will need `enable-nfs-ldap.yml` so that your diego-cell is configured correctly to use a LDAP server. If you deploy yours without `enable-nfs-ldap.yml` your deployment will still pass, but it's passing without the LDAP configuration on the diego-cell.

winkingturtle-vmw: The develop branch will correctly generate `ldap_test_server_ssl`, but it doesn't get used by `enable-nfs-ldap.yml` opsfile. The reason for this PR is to ensure CA Cert is used in the opsfile. If the opsfile uses a different CA cert than what ldap test server generates, we would have to have yet another another opsfile that will sync the two. 

The way I see it, we have three options:

1. If you don't like the new name for LDAP CA Cert `ldap_ca_cert` as you proposed, we can go back to the old name.
2. If we are okay with the name, this PR can be merged
3. We close this PR and leave develop branch as is and we can have an opsfile in our release to this very thing.

any other thoughts?

winkingturtle-vmw: > The test ops file will overwrite the "ldap_ca_cert" property with the test certificate. I've verified that with a bosh interpolate run.

I am not sure what are you referring to here. If the LDAP test server opsfile (`enable-nfs-test-ldapserver.yml`) can just generate `ldap_server_ssl` why would there be a need for an override?

jochenehret: Ok, if we choose the same name for the certificate, there would be no need for an overwrite. So we can name the certificate `ldap_server_ssl` instead of `ldap_ca_cert` (that name would be confusing because technically it's not a certificate authority). Then this PR is ready for merge!

winkingturtle-vmw: I've pushed a commit to reflect those changes

jochenehret (approved):

PR #1230, opened 2025-02-07
Remove experimental ops file disable-v2-api

dimivel:

API v2 is disabled by default, see https://github.com/cloudfoundry/capi-release/pull/503 and [RFC0032 - CF API v2 End of Life](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0032-cfapiv2-eol.md)

Since API v2 is disabled by default there is no need of experimental ops file disable-v2-api.

- [ ] YES
- [ ] NO

- [x] NO

> removes ops-files from the following folders
>    - `./operations/experimental`
>    

- [ ] YES - please specify
- [x] NO

- [ ] GA'd feature/component

> _Few examples_


operations/experimental/README.md
```diff
@@ -11,43 +11,43 @@ This is the README for Experimental Ops-files. To learn more about `cf-deploymen
 
 # Experimental ops files
 
-| Name | Purpose | Notes | Currently validated in Release Integration CI pipelines? |
```
jochenehret, line 14: Have you opened the Markdown file in IntelliJ? It reformats the whole file which makes diffs hard to detect. Better open it in a plain text editor.

jochenehret (changes requested):

operations/experimental/README.md
dimivel: I managed to adjust the file in sublime and the diff looks much better now.

jochenehret (dismissed):

jochenehret:
It's also not a breaking change as the ops file is not removed, can you please adapt the description?

PR #1232, opened 2025-02-12
Add jwt bearer for user authentication

strehle:

Add JWT bearer to the defaults of allowed grant type.
JWT bearer should replace password grants, but we need to have both for a while

A default CF should have it, because of
cloudfoundry/cli#3368
Another reason to allow alternative to password grant https://datatracker.ietf.org/doc/html/rfc9700#section-2.4

No problem, but enhancement. Customer should not use their password, but tokens, therefore support jwt bearer to support customer needs.

cloudfoundry/cli#3368

https://datatracker.ietf.org/doc/html/rfc9700#section-2.4

- [X] YES
- [ ] NO

- [X] NO

Yes please, annonce it as new possibility, if cloudfoundry/cli#3368 is merged.

- [ ] YES - please specify
- [X] NO

- [X] GA'd feature/component

JWT bearer with cf client works,

UAA Team

### DIFF
--- a/cf-deployment.yml
+++ b/cf-deployment.yml
@@ -660,7 +660,7 @@ instance_groups:
           cf:
             access-token-validity: 1200
             authorities: uaa.none
-            authorized-grant-types: password,refresh_token
+            authorized-grant-types: password,refresh_token,urn:ietf:params:oauth:grant-type:jwt-bearer
             override: true
             refresh-token-validity: 2592000
             scopes:


strehle:
@jochenehret FYI, this PR is more useful if https://github.com/cloudfoundry/cli/pull/3397 is accepted, because then the normal USER in cf can run it, but even without cf - client adoption the REST calls to UAA can be done

jochenehret:
Ok. Are there any possible side-effects we should be aware of?

strehle:
> Ok. Are there any possible side-effects we should be aware of?

Only this change has no effect (include side effect), but this change can improve situation if we have at least the enhancements in cf client.

I am not sure if we need a RFC from CF side to announce it.

The bigger change or breaking change will be if we finally remove password as authorized-grant-types from cf client. So I see it similar to move from v2 to v3. This change here simply adds v3 and therefore no side effect now.

jochenehret (approved):

PR #1255, opened 2025-07-09
Add security group for handling ipv6 traffic

Milena-Encheva:

It is pushed against develop branch and it is only a draft for discussion purposes. 

New security group allowing IPv6 traffic is added to the cf-deployment.yml

https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0038-ipv6-dual-stack-for-cf.md

- [ ] YES
- [X] NO

- [X] NO

>    - `./addons`
>

- [ ] YES - please specify
- [X] NO

- [ ] GA'd feature/component

> _Few examples_

@oliver-heinrich


operations/experimental/enable-ipv6-egress.yml
```diff
@@ -0,0 +1,27 @@
...
+  - public_networks
+  - dns
+  - ipv6_public_networks
+
+- type: replace
+  path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/cc/security_group_definitions/-
+  value:
+    name: ipv6_public_networks
+    rules:
+    - destination: 2000::-3fff:ffff:ffff:ffff:ffff:ffff:ffff:ffff
```
peanball, line 26: The `destination` also handles CIDR notation.
```suggestion
    - destination: 2000::/3
```

operations/experimental/add-ipv6-security-group.yml
```diff
@@ -0,0 +1,22 @@
...
+- type: replace
+  path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/cc/default_staging_security_groups
+  value:
+  - public_networks
+  - dns
+  - ipv6_public_networks
+
+- type: replace
+  path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/cc/security_group_definitions/-
+  value:
```
peanball, line 18: As discussed in a call, I would like to have a note _in_ the file that makes clear what you enable if you take this opsfile wholesale.
```suggestion
  value:
  # This is a sample IPv6 allowlist that allows the entirety of
  # the public IPv6 address space. Please note that this may include
  # the Diego cells and other nodes managed by BOSH, which might be
  # considered "internal" and are protected from external requests
  # via global firewall rules that apply for the CF foundation.
  # Please exclude any addresses from the range that should not be
  # accessible from app containers.
```

operations/experimental/add-ipv6-security-group.yml
```diff
@@ -0,0 +1,22 @@
+# --- Add IPv6 Security Groups ---
+- type: replace
```
jochenehret, line 2: You can use the "-" syntax also here:
```
- type: replace
  path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/cc/default_running_security_groups/-
  value: ipv6_public_networks
```
This adds the "ipv6_public_networks" at the end of the array. So you don't need to repeat the existing values and it interferes less with possible other custom ops files.

jochenehret (changes requested):
We also need a test for the new ops file: operations/experimental/add-ipv6-security-group.yml

peanball (commented):
Some editorial things, a naming preference and a bit of text warning about allowlisting the whole public IPv6 address range, and why we warn.

oliver-heinrich (approved):
LGTM

PR #1335, opened 2026-05-05
Adding an experimental ops file to enable cflinuxfs5.

AttilaAlmasi:
### WHAT is this change about?
As part of the validation process for the new cflinuxfs5 stack we introduce a new experimental ops file which will serve alongside the cflinuxfs4 stack. 

> _Please describe the change._
- Added new add-cflinuxfs5.yml ops file. j
- Updated the README.md 
- added the new file reference to the tests

### What customer problem is being addressed? Use customer persona to define the problem e.g. Alana is unable to...
Since cflinuxfs4 is based on Jammy which is end of life in 2027 we need to introduce the new cflinuxfs5 stack which is based on Ubuntu Noble.

### Please provide any contextual information.

> _Include any links to other PRs, stories, slack discussions, etc... that will help establish context._
https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0039-noble-based-cflinuxfs5.md 

### Has a cf-deployment including this change passed [cf-acceptance-tests](https://github.com/cloudfoundry/cf-acceptance-tests)?

- [ ] YES
- [x] NO

### Does this PR introduce a breaking change? Please take a moment to read through the examples before answering the question.

- [ ] YES - please choose the category from below. Feel free to provide additional details.
- [x] NO

> **Types of breaking changes:**
> 1. **causes app or operator downtime**
> 2. increases VM footprint of cf-deployment - e.g. new jobs, new add ons, increases # of instances etc.
> 3. modifies, deletes or moves the name of a job or instance group in the main manifest
> 4. modifies the name or deletes a property of a job or instance group in the main manifest
> 5. changes the name of credentials in the main manifest
> 6. requires out-of-band manual intervention on the part of the operator
> 7. modifies the ops-file path, changes the type, changes the values or removes ops-files from the following folders
>    - `./operations/` or `./operations/experimental`
>    - `./addons`
>    - `./backup-and-restore/`
>
> _If you're promoting an experimental Ops-file (or removing one), Please follow the [Ops-file workflows](https://github.com/cloudfoundry/cf-deployment/blob/main/ops-file-promotion-workflow.md)._

> Ops files changes in the following folders are considered as NON BREAKING CHANGES
> `./community`, `./example-vars-files`, `./test`

### How should this change be described in cf-deployment release notes?
Added experimental ops file for cflinuxfs5

[...]

Issue #1323, opened 2026-04-08
Integration of cflinuxfs5 stack


jochenehret:
### What is this issue about?

The [cflinuxfs5](https://github.com/cloudfoundry/cflinuxfs5) stack is now available, based on Ubuntu 24.04 "Noble Numbat". Application developers will need to move to this stack before cflinuxfs4 leaves support.

In this issue we want to track the integration of cflinuxfs5 into the cf-deployment release and validation process.

#### cflinuxfs5 stack
- [x] [cflinuxfs5-release](https://github.com/cloudfoundry/cflinuxfs5-release)

#### cflinuxfs5 support for buildpacks shipped with cf-deployment (according to release notes):
- [ ] [java-buildpack](https://github.com/cloudfoundry/java-buildpack)
- [ ] [dotnet-core-buildpack](https://github.com/cloudfoundry/dotnet-core-buildpack)
- [ ] [nodejs-buildpack](https://github.com/cloudfoundry/nodejs-buildpack)
- [ ] [python-buildpack](https://github.com/cloudfoundry/python-buildpack)
- [ ] [go-buildpack](https://github.com/cloudfoundry/go-buildpack)
- [ ] [ruby-buildpack](https://github.com/cloudfoundry/ruby-buildpack)
- [ ] [staticfile-buildpack](https://github.com/cloudfoundry/staticfile-buildpack)
- [ ] [php-buildpack](https://github.com/cloudfoundry/php-buildpack)
- [ ] [nginx-buildpack](https://github.com/cloudfoundry/nginx-buildpack)
- [ ] [r-buildpack](https://github.com/cloudfoundry/r-buildpack)
- [ ] [binary-buildpack](https://github.com/cloudfoundry/binary-buildpack)

#### Integration of cflinuxfs5 into CATs and cf-deployment
- [ ] experimental ops file to ship cflinuxfs5 and supported buildpacks as opt-in 
- [ ] experimental ops file to make cflinuxfs5 default stack
- [ ] run CATS against cflinuxfs5 (either in experimental-hermione or in a separate pipeline)
- [ ] promote cflinuxfs5 ops files (could be skipped if we integrate the ops files directly into cf-deployment.yml)
- [ ] rollout step 1 (make cflinuxfs5 the default)
- [ ] rollout step 2 (remove cflinuxfs4 leftovers)

oliver-heinrich:
Hi @jochenehret, 

does it make sense to split 'task'
[...]

PR #1339, opened 2026-05-11
Add cflinuxfs5 stack to experimental pipeline validation

Milena-Encheva:

Adds cflinuxfs5 stack validation to the experimental Concourse pipeline (hermione). This includes applying the add-cflinuxfs5.yml ops file in the experimental deploy jobs, and introducing a new experimental-cats-cflinuxfs5 job to run cf-acceptance-tests against the cflinuxfs5 stack.

George (CF operator) needs confidence that the upcoming cflinuxfs5 stack works correctly with cf-deployment before it is promoted as a default or GA'd stack. Without this pipeline validation, there is no automated signal that cflinuxfs5 is stable and compatible with current cf-deployment.

Part of https://github.com/cloudfoundry/cf-deployment/issues/1323. Follows the same experimental validation approach previously used for cflinuxfs4 before it became the default stack. Depends on operations/experimental/add-cflinuxfs5.yml (introduced in [Attila's PR](https://github.com/cloudfoundry/cf-deployment/pull/1335)) and the cflinuxfs5_integration_config.json in relint-envs.

- [ ] YES
- [X] NO

- [X] NO

Add cflinuxfs5-release to experimental pipeline validation; introduces experimental-cats-cflinuxfs5 job to run CATs against the cflinuxfs5 stack.

- [X] YES - cflinuxfs5-release is added to opsReleases in inputs.yml
- [ ] NO

- [ ] GA'd feature/component

1.) The experimental-deploy job in the hermione pipeline applies operations/experimental/add-cflinuxfs5.yml without errors.
2.) After a successful experimental-deploy, the new experimental-cats-cflinuxfs5 job triggers and completes CATs against the cflinuxfs5 stack using cflinuxfs5_integration_config.json.

@oliver-heinrich @jochenehret

### DIFF
--- a/ci/pipelines/cf-deployment.yml
+++ b/ci/pipelines/cf-deployment.yml
@@ -1344,12 +1344,12 @@ jobs:
     do:
     - get: experimental-pool
       trigger: true
-      passed: [ experimental-cats ]
+      passed: [ experimental-deploy ]
     - in_parallel:
       - get: cf-acceptance-tests-rc
       - get: relint-envs
       - get: cf-deployment-develop
-        passed: [ experimental-cats ]
+        passed: [ experimental-deploy ]
       - get: cf-deployment-concourse-tasks
     - task: update-integration-configs
       file: cf-deployment-concourse-tasks/update-integration-configs/task.yml
@@ -1368,10 +1368,10 @@ jobs:
         CONFIG_FILE_PATH: environments/test/hermione/cflinuxfs5_integration_config.json
         REPORTER_CONFIG_FILE_PATH: environments/test/hermione/reporter_config.json
         RELINT_VERBOSE_AUTH: "true"
+    NODES: 6
 
 - name: experimental-delete-deployment
   serial: true
-  serial_groups: [ experimental-cats-cflinuxfs5 ]
   public: true
   plan:
   - get: experimental-pool
--- a/ci/pipelines/cf-deployment.yml
+++ b/ci/pipelines/cf-deployment.yml
@@ -1335,6 +1335,7 @@ jobs:
         REPORTER_CONFIG_FILE_PATH: environments/test/hermione/reporter_config.json
         RELINT_VERBOSE_AUTH: "true"
         SKIP_REGEXP: "shows logs and metrics"
+        NODES: 6
 
 - name: experimental-cats-cflinuxfs5
   serial_groups: [ experimental-cats-cflinuxfs5 ]
@@ -1368,7 +1369,7 @@ jobs:
         CONFIG_FILE_PATH: environments/test/hermione/cflinuxfs5_integration_config.json
         REPORTER_CONFIG_FILE_PATH: environments/test/hermione/reporter_config.json
         RELINT_VERBOSE_AUTH: "true"
-    NODES: 6
+        NODES: 6
 
 - name: experimental-delete-deployment
   serial: true
--- a/ci/pipelines/cf-deployment.yml
+++ b/ci/pipelines/cf-deployment.yml
@@ -22,6 +22,7 @@ groups:
   - experimental-deploy
   - experimental-smoke-tests
   - experimental-cats
+  - experimental-cats-cflinuxfs5
   - experimental-delete-deployment
   - lite-acquire-pool
   - lite-deploy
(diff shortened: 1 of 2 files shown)


oliver-heinrich (approved):
LGTM

ci/pipelines/cf-deployment.yml
```diff
@@ -1332,6 +1336,39 @@ jobs:
...
+  - timeout: 4h
+    do:
+    - get: experimental-pool
+      trigger: true
+      passed: [ experimental-deploy ]
+    - in_parallel:
+      - get: cf-acceptance-tests-rc
+      - get: relint-envs
+      - get: cf-deployment-develop
+        passed: [ experimental-deploy ]
```
jochenehret, line 1353: This job should run after "experimental-cats". Otherwise, two CATs suites will run in parallel and might overload the deployment.

ci/pipelines/cf-deployment.yml
```diff
@@ -1340,6 +1377,7 @@ jobs:
     trigger: true
     passed:
     - experimental-cats
+    - experimental-cats-cflinuxfs5
```
jochenehret: The new job "experimental-cats-cflinuxfs5" will likely be red for some time. So the downstream delete job wouldn't be executed and the VMs wouldn't be cleaned up. Not sure how we solved this last time. One solution would be to execute the delete step in a Concourse "ensure" block: https://concourse-ci.org/docs/steps/modifier-and-hooks/ensure/

jochenehret (changes requested):

ci/pipelines/cf-deployment.yml
Milena-Encheva: Made it experimental-cats-cflinuxfs5 to start after the regular CATs job finishes.

ci/pipelines/cf-deployment.yml
Milena-Encheva: Removed experimental-cats-cflinuxfs5 from the passed constraint (so delete triggers after experimental-cats and experimental-smoke-tests pass), and added serial_groups: [ experimental-cats-cflinuxfs5 ] to both jobs. This way, I think Concourse will wait for any in-progress cflinuxfs5 run to finish before starting cleanup, without blocking cleanup entirely if cflinuxfs5 is permanently red.

jochenehret: Not sure if this really works. Concourse could start the delete job first, and then afterwards (when the "experimental-cats-cflinuxfs5" serial group is released again) it could try to run the "experimental-cats-cflinuxfs5" job. That one would fail as the cf deployment is now deleted.

Last time (for the cflinuxfs4 migration) we ran both CATs jobs in parallel:
https://github.com/cloudfoundry/cf-deployment/commit/9bf7e82b523e811adf91fe99423e7b21441ea47f
I think it worked reliably, but we should check the system load and reduce the number of nodes if necessary (default is 12: https://github.com/cloudfoundry/cf-deployment-concourse-tasks/blob/77f2c8ac6add085feba3df606967f320c9acc37d/run-cats/task.yml#L25)

The "delete" job was however always triggered when the regular CATs job finished (not waiting for the new cflinxufs4 CATs job). That could work if we reduce the number of test suites in the cflinuxfs5 job.

Milena-Encheva: Last changes: 

1.) experimental-cats-cflinuxfs5 triggers after experimental-deploy: Runs in parallel with experimental-cats to avoid blocking the pipeline.

2.) Added NODES: 6: Reduces load during parallel execution.

3.) Removed serial_groups from experimental-delete-deployment: Cleanup triggers after experimental-cats and experimental-smoke-tests, regardless of cflinuxfs5 status.

ci/pipelines/cf-deployment.yml
```diff
@@ -1332,6 +1336,40 @@ jobs:
...
+    - task: run-cats
+      input_mapping:
+        integration-config: updated-integration-configs
+        cf-acceptance-tests: cf-acceptance-tests-rc
+      file: cf-deployment-concourse-tasks/run-cats/task.yml
+      params:
+        CONFIG_FILE_PATH: environments/test/hermione/cflinuxfs5_integration_config.json
+        REPORTER_CONFIG_FILE_PATH: environments/test/hermione/reporter_config.json
+        RELINT_VERBOSE_AUTH: "true"
+    NODES: 6
```
jochenehret: Wrong indentation?

jochenehret (changes requested):
The regular "experimental-cats" job should also have `NODES: 6` configured.

jochenehret (approved):
Looks good, but note that you must first merge https://github.com/cloudfoundry/relint-envs/pull/101 before uploading the new pipeline.

PR #1252, opened 2025-07-02
Adds experimental support for MySQL 8.4

ryanwittrup:
As of PXC v1.0.0, MySQL 8.4 is available for opt-in support. This change is meant to be used for experimental testing and validation.

If everything goes well when tests are run in CI, we can make the permanent change to upgrade the default from 8.0 to 8.4

[TNZ-45485](https://vmw-jira.broadcom.net/browse/TNZ-45485)

Authored-by: Ryan Wittrup <ryan.wittrup@broadcom.com>

## Please take a moment to review the questions before submitting the PR
Noted

### WHAT is this change about?
[PXC v1.1.0 was recently released](https://github.com/cloudfoundry/pxc-release/releases/tag/v1.1.0), which includes support for MySQL 8.4. This is still an opt-in feature for now, with plans to move to 8.4 as the default in a future 2.X release.

For now, we want to start testing CF with the 8.4 version - if everything looks good in CI, we can open another change to make this the default version for the internal database.

### What customer problem is being addressed? Use customer persona to define the problem e.g. Alana is unable to...
MySQL 8.0 reaches EOL in early 2026, so platform operators should begin to make plans now to migrate to the long-term supported version

### Please provide any contextual information.
NA

### Has a cf-deployment including this change passed [cf-acceptance-tests](https://github.com/cloudfoundry/cf-acceptance-tests)?

- [x] YES
- [ ] NO

### Does this PR introduce a breaking change? Please take a moment to read through the examples before answering the question.

- [ ] YES - please choose the category from below. Feel free to provide additional details.
- [x] NO

> **Types of breaking changes:**
> 1. **causes app or operator downtime**
> 2. increases VM footprint of cf-deployment - e.g. new jobs, new add ons, increases # of instances etc.
> 3. modifies, deletes or moves the name of a job or instance group in the main manifest
> 4. modifies the name or deletes a property of a job or instance group in the main manifest
> 5. changes the name of credentials in the main manifest
> 6. requires out-of-band manual intervention on the part of the operator
> 7. modifies the ops-file path, changes the type, changes the values or removes ops-files from the following folders
>    - `./operations/` or `./operations/experimental`
>    - `./addons`
>    - `./backup-and-restore/`
>
> _If you're promoting an experimental Ops-file (or removing one), Please follow the [Ops-file workflows](https://github.com/cloudfoundry/cf-deployment/blob/main/ops-file-promotion-workflow.md)._

> Ops files changes in the following folders are considered as NON BREAKING CHANGES
[...]

PR #1343, opened 2026-05-26
Feat: Use-mysql-version-8.4-by-default-in-cfd

iaftab-alam:

This PR aims to change the default mysql_version in cf-D from 8.0 to 8.4 and deprecate the experimental/use-mysql-version-8.4.yml ops file.

The ops file has been validated in the Release Integration CI pipelines (hermione environment) and is now promoted inline into the base manifest per the ops-file promotion workflow.

Alana expects the default MySQL version used to be in-support.

- MySQL version 8.0 is EOL.
- https://github.com/cloudfoundry/cf-deployment/pull/1252

- [ ] YES
- [X] NO

- [X] YES - CF operators who have not already used the provided ops file to upgrade to MySQL version 8.0 will find their CF-D forcibly upgraded!
- [ ] NO

This release changes the default MySQL version 8.0 to 8.4 (Percona XtraDB Cluster). Deploying directly onto an existing MySQL 8.0 environment without following the [PXC in-place upgrade procedure](https://github.com/cloudfoundry/pxc-release#deploying-pxc-release-with-percona-xtradb-cluster-84-support) might result in a failed deployment.

Operators currently using use-mysql-version-8.4.yml are unaffected and can safely remove it from their pipelines — it is now a no-op.

- [ ] YES - please specify
- [X] NO

- [X] GA'd feature/component

- The pxc release should make use MySQL version 8.4!
- New deployments should work without any issue.

None


units/tests/experimental_test/operations.yml
```diff
@@ -69,10 +69,6 @@ use-trusted-ca-cert-for-apps-cflinuxfs4.yml:
     - add-cflinuxfs4.yml
   varsfiles:
     - ../example-vars-files/vars-use-trusted-ca-cert-for-apps.yml
-use-mysql-version-8.4.yml:
```
jochenehret, line 72: I think you need the unit test if the ops file is still present?

iaftab-alam: Thanks Jochen for feedback! You're right! the I thought the test entry should stay removed! but testing locally it fails. The ops file content is now pure comments.

Validation locally:
There's one failure: ensure_use-mysql-version-8.4.yml_has_test_coverage. So this means the test framework requires every ops file present in the folder to have an entry in operations.yml even for deprecated ones!

I will adjust the changes!

iaftab-alam: It is ready for a review again!

oliver-heinrich (approved):
lgtm

jochenehret (approved):

PR #1345, opened 2026-05-28
Fix compatibility between disable-cf-credhub and add-cflinuxfs5 ops f…

Milena-Encheva:

Adds a `remove` operation in `disable-cf-credhub.yml` to strip the `((credhub_tls.ca))` reference from the `cflinuxfs5-rootfs-setup` job's `trusted_certs`, mirroring the existing removal already done for `cflinuxfs4-rootfs-setup`.

Cody is unable to deploy CF when combining the `add-cflinuxfs5.yml` and `disable-cf-credhub.yml` ops files. The deployment fails at render time because `cflinuxfs5-rootfs-setup` still references `((credhub_tls.ca))` after the `credhub_tls` variable has been removed by `disable-cf-credhub.yml`.

- Pipeline failure: https://concourse.wg-ard.ci.cloudfoundry.org/teams/main/pipelines/cf-deployment/jobs/experimental-deploy/builds/2430
- `add-cflinuxfs5.yml` adds `((credhub_tls.ca))` to `cflinuxfs5-rootfs-setup` trusted certs (index 1), but `disable-cf-credhub.yml` was only cleaning up that reference for `cflinuxfs4-rootfs-setup`.
- The `?` suffix on the new path makes the operation a no-op when `add-cflinuxfs5.yml` has not been applied.

- [ ] YES
- [x] NO

- [ ] YES
- [x] NO

Cody can now deploy CF using both `add-cflinuxfs5.yml` and `disable-cf-credhub.yml` ops files together without a render failure caused by a dangling `((credhub_tls.ca))` reference in the `cflinuxfs5-rootfs-setup` job.

- [ ] YES
- [x] NO

- [ ] GA'd feature/component

Experimental-deploy is not failing.
The deployment should complete without the error:
> `Failed to find variable '.../credhub_tls' from config server: HTTP Code '404'`

- [x] **Urgent** - unblocks current or future work (fixes the `experimental-deploy` pipeline)

@oliver-heinrich @jochenehret


jochenehret (changes requested):
That's not working for the default case:
```
bosh interpolate cf-deployment.yml -o operations/experimental/disable-cf-credhub.yml

Error 'operation [7] in operations/experimental/disable-cf-credhub.yml failed': Expected to find exactly one matching array item for path '/instance_groups/name=diego-cell/jobs/name=cflinuxfs5-rootfs-setup' but found 0

Exit code 1
```
You can also check with the unit tests:
```
go test ./tests/experimental_test/...
--- FAIL: TestExperimental (0.00s)
    --- FAIL: TestExperimental/disable-cf-credhub.yml (0.04s)
        operations_test.go:21: interpolate failed:
            Error 'operation [7] in disable-cf-credhub.yml failed': Expected to find exactly one matching array item for path '/instance_groups/name=diego-cell/jobs/name=cflinuxfs5-rootfs-setup' but found 0
FAIL
FAIL	github.com/cf-deployment/units/tests/experimental_test	46.056s
FAIL
```
Unfortunately, the units tests don't run automatically for a new PR (only after someone approved it).

oliver-heinrich (approved):
Can we make the ops file configurable for the cflinuxfs[4|5] part in path:
path: /instance_groups/name=diego-cell/jobs/name=cflinuxfs5-rootfs-setup/properties/cflinuxfs5-rootfs/trusted_certs/1 
dependent which stack is used? We may introduce an ops file dependency where 'add_cflinuxfs5' ops file must be first deployed.
Alternatively we may need another 'disable_cf_credhub_fs5' ops file. Which is also not a nice solution from my point of view. 
Any further ideas?

Milena-Encheva:
Added the` ?` after the flinuxfs5-rootfs-setup , not at the end of the line. The other option we discussed with @iaftab-alam was to change the order of OPs files, but in this case this should work too.

jochenehret (approved):
Approving, please wait for unit tests to pass before merging!

PR #1359, opened 2026-07-09
Add kind-deployment validation to CI pipeline

Milena-Encheva:

This PR targets develop. ✅ 

 Adds Cloud Foundry on Kubernetes (kind) validation to the cf-deployment Concourse pipeline. Two new jobs — kind-smoke-tests and kind-cats — dispatch the existing kind-smoke.yaml and kind-cats.yaml GitHub Actions workflows in cloudfoundry/kind-deployment with the fresh-validation: true input. That input (added in [kind-deployment#424](https://github.com/cloudfoundry/kind-deployment/pull/424), already merged) makes the workflow sync its Helm chart versions from cf-deployment's develop branch before deploying, so the Kubernetes deployment is tested with the same release versions currently being validated on BOSH.

The reusable trigger-github-workflow task has been moved to [runtime-ci#687](https://github.com/cloudfoundry/runtime-ci/pull/687) so it can be shared across pipelines. The gh CLI required by the task is provided by the cloudfoundry/cf-deployment-concourse-tasks image — see [cf-deployment-concourse-tasks#249 ](https://github.com/cloudfoundry/cf-deployment-concourse-tasks/pull/249) which adds it to the image.

Alana, a CF release engineer, is unable to know whether the component versions on develop — already validated on BOSH — also deploy and pass acceptance tests on Kubernetes. Today kind-deployment only syncs from tagged cf-deployment releases, which lag develop by days or weeks, so there is no early signal that a version bump breaks the Kubernetes path. This change gives that signal at the same time BOSH validation runs.

- The kind jobs trigger on the same gate as every BOSH environment (unit-test-ops-files, lint-cf-deployment-manifest, unit-test-golang-tests, unit-test-update-releases-coverage), so they run in parallel with fresh/upgrade/experimental/windows/lite/FIPS.
- They are grouped under a new kind-neville pipeline group (Harry Potter naming convention).
- The version sync now happens inside the kind workflows via fresh-validation; no separate sync job or PR branch is needed.
- The trigger-github-workflow task lives in  [runtime-ci#687](https://github.com/cloudfoundry/runtime-ci/pull/687)  and is referenced via the existing runtime-ci resource. It uses the cloudfoundry/cf-deployment-concourse-tasks image (which gains the gh CLI via [cf-deployment-concourse-tasks#249 ](https://github.com/cloudfoundry/cf-deployment-concourse-tasks/pull/249).

- The jobs are non-blocking — bless-manifest does not depend on them, so a kind failure will not gate releases while the Kubernetes path matures.

- [ ] YES
- [X] NO - this is a CI pipeline change only; it does not modify the manifest or any BOSH release. It does not alter what CATS runs against. (The kind path itself runs CATS on Kubernetes.)

- [X] NO - CI configuration only.

>    - `./addons`
>

Not release-noteworthy — internal CI change with no operator-facing impact. (Suggested, if included: "CI now validates develop versions on Kubernetes via kind-deployment, in parallel with BOSH.")

- [ ] YES - please specify
- [X] NO

- [ ] GA'd feature/component
- Neither — no manifest/ops-file feature change. This is CI tooling only.

- kind-smoke-tests and kind-cats appear in the pipeline under the kind-neville group.
- Both trigger automatically after the four unit/lint jobs pass on develop.
- Each dispatches the correct kind-deployment workflow on main with fresh-validation: true and watches it to completion, reflecting pass/fail in Concourse.
- bless-manifest continues to pass/fail independently of the kind jobs.
- Pipeline YAML validates and fly set-pipeline applies cleanly.
- Depends on [cf-deployment-concourse-tasks#249 ](https://github.com/cloudfoundry/cf-deployment-concourse-tasks/pull/249)  and  [runtime-ci#687](https://github.com/cloudfoundry/runtime-ci/pull/687) being merged first.
 

@oliver-heinrich @jochenehret


ci/tasks/trigger-github-workflow/task.sh
```diff
@@ -0,0 +1,102 @@
...
+# Requires these environment variables:
+#   GITHUB_TOKEN      – PAT or GitHub App token with Actions scope
+#   GITHUB_REPO       – owner/repo  (e.g. cloudfoundry/kind-deployment)
+#   WORKFLOW_FILE     – workflow filename  (e.g. kind-smoke.yaml)
+#   WORKFLOW_REF      – git ref to run against  (default: main)
+#   WORKFLOW_INPUTS   – JSON object of workflow inputs  (default: {})
+##
+
+# ── install gh CLI and jq ───────────────────────────────────────────
+apk add --no-cache --quiet \
```
jochenehret: CLI tools could be added to the cf-deployment-concourse-tasks image:
https://github.com/cloudfoundry/cf-deployment-concourse-tasks/blob/f5e2b103800315eac9fa6ba3e7b6c6054e17816c/ci/pipeline.yml#L12

ci/tasks/trigger-github-workflow/task.yml
```diff
@@ -0,0 +1,17 @@
+---
```
jochenehret: Special Concourse tasks can be stored in the runtime-ci repository:
https://github.com/cloudfoundry/runtime-ci

ci/tasks/trigger-github-workflow/task.yml
```diff
@@ -0,0 +1,17 @@
...
+platform: linux
+image_resource:
+  type: docker-image
```
jochenehret: You can re-use the cf-deployment-concourse-tasks image here.

ci/pipelines/kind-neville.md
```diff
@@ -0,0 +1,374 @@
...
+- **Renovate** – independently watches upstream GitHub releases for each
+  component and creates per-component PRs.
+
+### The gap
+
+The existing daily sync in kind-deployment is **batch and delayed**: it only
+picks up versions after they appear in a cf-deployment *release* (tagged),
+which can be days or weeks after they land on `develop`. There is no
+automated path that validates whether the versions currently being tested in
+BOSH environments also work on Kubernetes.
```
c0d1ngm0nk3y: Probably I haven't understood it, but was is the gap? The `main` in `kind-deployment` uses the versions validated by `cf-depoyment`. So every user get working version.

For validation, we introduced the parameter `fresh validation` which then uses the versions from the `develop` branch in `cf-deployment` instead.

The helm charts and docker files are automatically created when a new bosh release (capi, diego, ...) is created. Independently if `cf-deployment` is already using the version or not yet.

If you still see a gap, please let me know that we can work on closing it :)

ci/pipelines/cf-deployment.yml
```diff
@@ -2131,6 +2180,52 @@ jobs:
...
+      passed:
+      - unit-test-ops-files
+      - lint-cf-deployment-manifest
+      - unit-test-golang-tests
+      - unit-test-update-releases-coverage
+    - get: runtime-ci
+  - task: trigger-kind-smoke-workflow
+    file: runtime-ci/tasks/trigger-github-workflow/task.yml
+    params:
+      GITHUB_TOKEN: ((ard_wg_gitbot_token))
```
c0d1ngm0nk3y: I created a dedicated token with the exact permissions you need. Please check the secret manager of the wg.

ci/pipelines/kind-neville.md
```diff
@@ -0,0 +1,374 @@
...
+ephemeral checkout and is discarded when the run finishes.
+
+---
+
+## Risks and Potential Problems
+
+### 1. Missing container images
+
+When cf-deployment `develop` has a new release version, the corresponding
+container image may not yet be built and pushed to `ghcr.io`. The sync
```
c0d1ngm0nk3y: It is not really a sync script checking for images and helm charts. What happens is this:

* A new github release is created for a bos release repo (e.g. capi-release)
* `renovate` will bump the version in `cf-k8s-releases` (once a day)
* The PR from renovate will be verified and automarged if successful
* In `main`, helm chart and docker images will be created and published

There can be 2 cases in which some artifact is missing:

* `cf-deployment` is too fast. `renovate` runs once a day and verification/release might take 1-2 hours.
* The artifact could not be created automatically (e.g. verification failed)

ci/pipelines/kind-neville.md
```diff
@@ -0,0 +1,374 @@
...
+**Mitigation:** The sync script reports which images are missing. Container
+images are typically built by kind-deployment's Renovate-triggered image
+build when version bumps are detected. There may be a gap where cf-deployment
+bumps a version before kind-deployment's image build runs. This is the most
+likely source of transient failures. Because kind is non-blocking for
+`bless-manifest`, such a failure will not hold up releases.
+
+### 2. Concourse-to-GitHub API reliability
+
+The trigger-github-workflow task depends on the GitHub API being available
```
c0d1ngm0nk3y: At least rate-limit shouldn't be a problem since you use a token.

ci/pipelines/kind-neville.md
```diff
@@ -0,0 +1,374 @@
...
+GitHub App token instead of a PAT.
+
+### 3. Serial group throughput
+
+Both kind jobs are in the same serial group, so smoke and CATS run
+sequentially rather than concurrently. A full cycle is roughly the sum of
+both durations. If the pipeline triggers frequently, kind validation may
+fall behind BOSH.
+
+**Mitigation:** Because smoke and CATS create independent kind clusters,
```
c0d1ngm0nk3y: An other mitigation would be to only use cats. I think it cannot happen that smoke tests find an error that cats doesn't find, right? It finds it much faster of course.

ci/pipelines/kind-neville.md
```diff
@@ -0,0 +1,374 @@
...
+heavy clusters on shared GitHub runners at once.
+
+### 4. Develop ref may advance between dispatch and sync
+
+`fresh-validation` syncs from `develop` at the moment the workflow runs.
+Additional commits may land on `develop` between the Concourse trigger and
+the in-workflow sync, so the test could validate a slightly newer or older
+set of versions than the exact commit that triggered the job.
+
+**Mitigation:** Acceptable — the kind check is informational and
```
c0d1ngm0nk3y: Another mitigation would be that we add a parameter to receive the exact commit sha that should be used. If that ever turn out to be a problem.

ci/pipelines/kind-neville.md
Milena-Encheva: You're right — "gap" was misleading. kind-deployment's main already uses validated versions. Updated the doc to remove the "gap" framing.

ci/pipelines/kind-neville.md
Milena-Encheva: Agreed, removed that concern from the doc since we use an authenticated token.

ci/pipelines/kind-neville.md
jochenehret: Well that depends on the CATs and smoke-tests configuration. smoke-tests could find issues if tests are enabled that are disabled in CATs.

jochenehret (approved):

